Issue #115, opened 2021-08-25
Add unit tests for > 2 arms trial


nociale:

gowerc:
Still need to add test cases to cover the "jackknife" scenario

PR #143, opened 2021-09-01
Test multiarm

gowerc:
Partially covers #115 

I haven't covered the case of only 1 arm. 

This PR did raise an interesting question for me though which is should we always be adding ivars$group into the MMRM model. Surely its plausible that the variable we create our reference groups by isn't actually a prognostic covariate in the model ?

### DIFF
--- a/tests/testthat/helper-misc.R
+++ b/tests/testthat/helper-misc.R
@@ -33,11 +33,17 @@ strip_names <- function(x) {
     return(x)
 }
 
+
 trunctate <- function(x, n) {
     floor(x * 10 ^ n) / 10 ^ n
 }
 
 
+#
+#         ** DEPRECIATED **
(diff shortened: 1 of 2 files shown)


nociale:
@gowerc As discussed here we need to try some experiments specifically for `jackknife` (also Marcel told me he might try something related to this). So, it would be nice if this branch would cover > 2 arms trial for `condmean/bootstrap`, `bayes` and `approxbayes`. Also a 2 arm trial with a subset of treatment or control arm as reference would be of interest (but I guess it is equivalent for our implementation expect the `analysis` part?) 

Additionally we need to start verifying that the package works for time-varying covariates, but I am confident it will work.

Do you prefer to merge this in the meanwhile?

gowerc:
I will merge this but then update the issue to indicate that more is still required

PR #156, opened 2021-09-15
Doc draws

nociale:
@gowerc It would be great of you could have a look at the documentation of `draws`, given its importance. I tried to summarize all the necessary information without being too long. Please feel free to comment about both minor and bigger issues. I have just one question: I wrote that `samples[[i]]$ids` contains the ids from the original dataset, while `samples[[i]]$ids_samp` contains the ids from the i-th sample. Is that correct?


R/draws.R
```diff
@@ -1,10 +1,112 @@
...
+#' according to the given multiple imputation methodology and returns the parameter
+#' estimates from the fit.
 #'
 #' @name draws
-#' @param data TODO
-#' @param data_ice TODO
-#' @param vars TODO
-#' @param method TODO
+#' @param data A `data.frame` containing the data to be used in the model.
+#' Must contain the outcome variable. See details.
```
gowerc, line 9: and any covariates ?

R/draws.R
```diff
@@ -1,10 +1,112 @@
...
+#' @param data_ice A `data.frame` that specifies the information related
+#' to the intercurrent events (ICE) and the imputation strategies. Its structure is described in the details.
```
gowerc, line 11: Perhaps just "See Details"

R/draws.R
```diff
@@ -1,10 +1,112 @@
...
+#' @param vars A `vars` object as generated by [set_vars()]. See details.
+#' @param method An object returned by the functions [method_bayes()],
```
gowerc, line 13: A `method` object as generated by either x , y or z. This specifies the ...

R/draws.R
```diff
@@ -1,10 +1,112 @@
-#' TODO
+#' @title Fit the base imputation model and get parameter estimates
+#'
+#' @description Assuming MAR, `draws` fits the base imputation model to the observed data
```
gowerc, line 3: I would start with the assumption perhaps:

> Fits the base imputation model to the observed data only using the observations that can be assumed to be MAR

or something like that ?

R/draws.R
```diff
@@ -1,10 +1,112 @@
...
+#' for the imputation phase (i.e. the regression coefficients and the covariance matrices
+#' from a MMRM model).
+#'
+#' The function distinguishes between the following methods:
+#' - Bayesian MI based on MCMC sampling: `draws` returns the draws
+#' from the posterior distribution of the parameters using a Bayesian approach based on
+#' MCMC sampling. This method can be adopted by specifying `method` using [method_bayes()].
+#' - Approximate Bayesian MI based on bootstrap: `draws` returns
+#' the draws from the posterior distribution of the parameters using a Bayesian approach,
+#' where the sampling from the posterior distribution is simulated fitting the MMRM model
```
gowerc, line 30: is simulated fitting - > is simulated by fitting

R/draws.R
```diff
@@ -1,10 +1,112 @@
...
+#' on bootstrap samples of the original dataset. This method can be adopted by
+#' specifying `method` through [method_approxbayes()].
+#' - Conditional mean imputation with bootstrap re-sampling: `draws` returns the
+#' ML/REML parameter estimates from the original dataset and from `n_samples` bootstrap samples.
```
gowerc, line 34: Do we need to say ML/REML as we didn't specify this in the approx bayes approach ?

R/draws.R
```diff
@@ -1,10 +1,112 @@
...
+#' This method can be adopted by specifying `method` through [method_condmean()] with
```
gowerc, line 35: (pedantic comment) not that keen on the word adopted. Perhaps

> This method can be specified by using `method = method_condmean()`

R/draws.R
```diff
@@ -1,10 +1,112 @@
...
+#' argument `type = "bootstrap"`.
+#' - Conditional mean imputation with Jackknife re-sampling: `draws` returns the
+#' ML/REML parameter estimates from the original dataset and from the leave-one-out samples.
```
gowerc, line 38: > and from --each--  leave-one-out sample 

?

R/draws.R
```diff
@@ -1,10 +1,112 @@
...
+#' outcome data must be set as `NA` instead of having the related row missing. Missing values
+#' in the covariates are not allowed. If `data` is incomplete
+#' then the [expand_locf()] helper function can be used to add in any missing rows, using
+#' Last Observation Carried Forward (LOCF) imputation to impute the covariates values.
+#'
+#' The argument `data_ice` contains information about the occurrence of ICEs. It is a
+#' `data.frame` with 3 columns:
+#' - A vector of characters related to the ids of the subjects that experienced
+#'   the ICE (must be named as `vars$subjid`).
+#' - A vector of levels of `data[[vars$visit]]` including the first visit after the ICE (must be named as `vars$visit`). If
```
gowerc, line 52: I would potentially start with a more basic level description  i.e.  The visit on which the ICE occurred on, this value must be equal to one of the levels of data[[vars...

R/draws.R
```diff
@@ -1,10 +1,112 @@
...
+#'   Possible imputation strategies are:
+#'   - `"MAR"`: Missing At Random.
+#'   - `"JTR"`: Jump To Reference.
+#'   - `"CIR"`: Copy Increments in Reference.
+#'   - `"CR"`: Copy Reference.
+#'   - `"JR"`: Jump to Reference.
+#'   - `"LMCF"`: Last Mean Carried Forward.
+#'   - User-defined imputation strategies can also be set.
+#'
+#' This argument is necessary at this stage since the model will be fit only on those observed data
```
gowerc, line 64: I am thinking that this might be easier to explain from the inverse  i.e.  

> The model is fitted onto the data after removing observation that occur after a non-MAR ICE.

R/draws.R
```diff
@@ -1,10 +1,112 @@
...
+#' that, if missing, would have been characterized as MAR. Thus, any non-MAR observed data, i.e.
+#' any observed data after `data_ice[[vars$visit]]` that is addressed with an imputation
+#' strategy different from MAR will be discarded for the model fit. However such observed data
+#' will not be discarded from the data in the imputation phase
+#' (performed with the function ([impute()]). To summarize, **at this stage only pre-ICE data
+#' and MAR post-ICE data are used**.
+#'
+#' The argument `vars` is a named list of key variable names to specify
```
gowerc, line 72: Perhaps something like

> The `vars` argument is a named list that specifies the names of key variables within `data` and `data_ice`.  Please use `set_vars()`. In particular the following named elements are used:

R/draws.R
```diff
@@ -1,10 +1,112 @@
...
+#' the imputation model. Can be set using [set_vars()] providing the following arguments:
+#' - `subjid`: column in `data` related to the subject ids variable.
```
gowerc, line 74: Not keen on the use of `related`

perhaps

> Name of the column in `data` and `data_ice` which contains the subject id

R/draws.R
```diff
@@ -1,10 +1,112 @@
...
+#' - `visit`: column in `data` related to the visit variable.
+#' - `outcome`: column in `data` related to the outcome variable.
+#' - `group`: column in `data` related to the group variable.
+#' - `covariates`: vector of characters related to the covariates to be included
+#'   in the model (including interactions).
+#'   By default the model specification is `outcome ~ 1 + visit + group`.
+#'   Please note that the `group*visit` interaction
+#'   is **not** included in the model by default.
+#' - `strata`: covariates used as stratification variables in the bootstrap.
+#'   By default only the `vars$group` is used as stratification variable.
```
gowerc, line 84: Potentially pedantic but I would pefer `set to`  instead of `used` as it might imply that we use group regardless of what the user specifies here.

gowerc (changes requested):

R/draws.R
nociale: I wanted to explicitly write that it should contains also the outcome variable. But yes, I think the necessary variables are subject ids, group, visit, outcome. Correct?

R/draws.R
nociale: Fits the base imputation model to the observed data assuming MAR? I just want to specify what we mean by "assuming MAR" in details later.

R/draws.R
nociale: Agreed. However "non-MAR ICE" is not well said, I should say something like "The model is fitted onto the data after removing the observations addressed with an imputation strategy different from MAR"

nociale: ok, let me try with this:

"This argument is necessary at this stage since the model is fitted after removing the observations whose underlying assumption is different from MAR. Thus, any observed data after `data_ice[[vars$visit]]` that is addressed with an imputation strategy different from MAR is discarded for the model fit. However such observations will not be discarded from the data in the imputation phase (performed with the function ([impute()]). To summarize, **at this stage only pre-ICE data and MAR post-ICE data are used**."

R/draws.R
```diff
@@ -1,10 +1,112 @@
...
+#' any observed data after `data_ice[[vars$visit]]` that is addressed with an imputation
+#' strategy different from MAR will be discarded for the model fit. However such observed data
+#' will not be discarded from the data in the imputation phase
+#' (performed with the function ([impute()]). To summarize, **at this stage only pre-ICE data
+#' and MAR post-ICE data are used**.
+#'
+#' The argument `vars` is a named list of key variable names to specify
+#' the imputation model. Can be set using [set_vars()] providing the following arguments:
+#' - `subjid`: column in `data` related to the subject ids variable.
+#' - `visit`: column in `data` related to the visit variable.
```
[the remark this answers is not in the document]
nociale: done

R/draws.R
nociale: done

R/draws.R
nociale: done

R/draws.R
nociale: done

R/draws.R
nociale: done

R/draws.R
nociale: done

R/draws.R
nociale: done

R/draws.R
nociale: I will use "MMRM" instead of "ML/REML"

R/draws.R
nociale: done

R/draws.R
nociale: done

R/draws.R
```diff
@@ -1,10 +1,112 @@
...
+#' in the covariates are not allowed. If `data` is incomplete
+#' then the [expand_locf()] helper function can be used to add in any missing rows, using
+#' Last Observation Carried Forward (LOCF) imputation to impute the covariates values.
+#'
+#' The argument `data_ice` contains information about the occurrence of ICEs. It is a
+#' `data.frame` with 3 columns:
+#' - A vector of characters related to the ids of the subjects that experienced
+#'   the ICE (must be named as `vars$subjid`).
+#' - A vector of levels of `data[[vars$visit]]` including the first visit after the ICE (must be named as `vars$visit`). If
+#'   multiple ICEs happen for the same subject, then only the first non-MAR visit should be set.
```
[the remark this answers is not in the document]
nociale: done

R/draws.R
nociale: done

R/draws.R
nociale: I will just omit "Must contain the outcome variable." since everything is described in the details

nociale:
@gowerc Sorry, I updated here a commit related to the validation of `index_mar`. I thought the best way of checking it was directly in the `longData` object, everytime the information related to `data_ice` is included in there (
i.e. within the `set_strategies` function). Could you please check if you like this idea and if the check stands in the right place of the code? Thanks!

gowerc:
@nociale ,  With regards to the validation function I wouldn't include it as a method directly inside of longData. Instead I would create it as a method of validate and then just assign a class on the fly when you call it i.e.

```
validate.index_mar <- function(x, ...){
    <logic here>
}
```

Then in your call to validate:

```
validate(  as_class(index_mar,   "index_mar"))
```

R/draws.R
gowerc: I think I like that, a few minor tweaks:
discarded -> excluded
 MAR post-ICE data  ->  post-ICE data that is MAR

R/draws.R
gowerc: Is "assumed" right, its more that it only uses MAR data rather than makes the assumption ?

R/draws.R
gowerc: Sure

R/draws.R
nociale: I guess we are assuming MAR, meaning that we fit the model assuming observed data are MAR.

From Cro et Al:
"..assuming MAR, fit a MVN distribution with an unstructured mean (ie, a separate mean for each of the baseline and post-randomization observation times) and variance-covariance matrix using a Bayesian approach.."

R/draws.R
nociale: ok!

R/draws.R
```diff
@@ -1,10 +1,97 @@
...
+#' argument `type = "bootstrap"`.
+#' - Conditional mean imputation with Jackknife re-sampling: `draws` returns the
+#' MMRM parameter estimates from the original dataset and from each leave-one-out sample.
+#' This method can be specified by using `method = method_condmean()` with
+#' argument `type = "jackknife"`.
+#'
+#' The argument `data` must have one row per visit per subject. This means that incomplete
+#' outcome data must be set as `NA` instead of having the related row missing. Missing values
+#' in the covariates are not allowed. If `data` is incomplete
+#' then the [expand_locf()] helper function can be used to add in any missing rows, using
```
gowerc, line 43: missing rows, using  ->  missing rows which uses

nociale: ok

R/draws.R
```diff
@@ -1,10 +1,97 @@
...
+#' Last Observation Carried Forward (LOCF) imputation to impute the covariates values.
+#'
+#' The argument `data_ice` contains information about the occurrence of ICEs. It is a
+#' `data.frame` with 3 columns:
+#' - A vector of characters related to the ids of the subjects that experienced
```
gowerc, line 48: Same as my previous comment I would remove the word  "related" and state more clearly what these variables represent. Perhaps something like: 

The argument `data_ice` contains information about the occurrence of ICEs. It is a `data.frame` with 3 columns:
- Subject ID - a character vector containing the subject ID for which the corresponding ICE belongs to. This column must be named as specified in `vars$subjid`
- Visit - A character vector container the visit in which the ICE occoured at. This column must be named as specified in `vars$visit`
- Strategy - A chacater vector specifying which imputation strategy should be used for this subject. This column must be named as specified in `vars$strategy` possible imputation strategies are: ...

R/draws.R
```diff
@@ -1,10 +1,97 @@
...
+#' base imputation model. The goal is to estimate the parameters of interest needed
+#' for the imputation phase (i.e. the regression coefficients and the covariance matrices
+#' from a MMRM model).
+#'
+#' The function distinguishes between the following methods:
+#' - Bayesian MI based on MCMC sampling: `draws` returns the draws
+#' from the posterior distribution of the parameters using a Bayesian approach based on
+#' MCMC sampling. This method can be specified by using `method = method_bayes()`.
+#' - Approximate Bayesian MI based on bootstrap: `draws` returns
+#' the draws from the posterior distribution of the parameters using a Bayesian approach,
```
gowerc, line 28: Is "using bayesian approach" potentially confusing here ?

R/draws.R
```diff
@@ -1,10 +1,97 @@
...
+#'   If multiple ICEs happen for the same subject, then only the first non-MAR visit should be used.
+#' - A vector of characters related to the imputation strategy to address the ICE (must be named as `vars$strategy`).
+#'   Possible imputation strategies are:
+#'   - `"MAR"`: Missing At Random.
+#'   - `"JTR"`: Jump To Reference.
+#'   - `"CIR"`: Copy Increments in Reference.
+#'   - `"CR"`: Copy Reference.
+#'   - `"JR"`: Jump to Reference.
+#'   - `"LMCF"`: Last Mean Carried Forward.
+#'   - User-defined imputation strategies can also be set.
```
gowerc, line 61: (pedantic) I would have this as a side remark rather than a bullet point. As in like afterwards "Please note that user-defined ..."

R/draws.R
```diff
@@ -1,10 +1,97 @@
...
+#'
+#' This argument is necessary at this stage since the model is fitted
```
gowerc, line 63: Given the paragraph break I would perhaps start with something like:

"To clarify the `data_ice` argument is necessary at this stage ..."

R/draws.R
```diff
@@ -1,10 +1,97 @@
...
+#' after removing the observations whose underlying assumption is different from MAR.
+#' Thus, any observed data after `data_ice[[vars$visit]]` that is addressed with an imputation
```
gowerc, line 65: should this  be   "after" -> "on or after"

R/draws.R
```diff
@@ -1,10 +1,97 @@
...
+#' strategy different from MAR is excluded for the model fit. However such observations
+#' will not be discarded from the data in the imputation phase
+#' (performed with the function ([impute()]). To summarize, **at this stage only pre-ICE data
+#' and post-ICE data that is MAR are used**.
+#'
+#' The vars argument is a named list that specifies the names of key variables within
+#' `data` and `data_ice`. Please use [set_vars()]. In particular the following named elements are used:
```
gowerc, line 72: "Please use [set_vars()]. In particular the following named elements are used:" 

Perhaps:

"This list is created by [set_vars()] and contains the following named elements:"

R/draws.R
nociale: Sure, with very little changes from your version I just addressed this.

R/draws.R
nociale: ok

R/draws.R
nociale: ok

R/draws.R
nociale: agreed

R/draws.R
nociale: I can use "using an approximate Bayesian approach"?

R/draws.R
nociale: ok

R/draws.R
```diff
@@ -1,10 +1,97 @@
...
+#' - `subjid`: name of the column in `data` and `data_ice` which contains the subject ids variable.
+#' - `visit`: name of the column in `data` and `data_ice` which contains the visit variable.
+#' - `outcome`: name of the column in `data` which contains the outcome variable.
+#' - `group`: name of the column in `data` which contains the group variable.
+#' - `covariates`: vector of characters which contains the covariates to be included
+#'   in the model (including interactions).
+#'   By default the model specification is `outcome ~ 1 + visit + group`.
```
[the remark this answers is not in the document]
nociale: agreed

R/draws.R
gowerc: sounds good to me

R/draws.R
gowerc: fair enough :)

R/draws.R
```diff
@@ -1,10 +1,97 @@
-#' TODO
+#' @title Fit the base imputation model and get parameter estimates
+#'
+#' @description fits the base imputation model to the observed data assuming MAR
```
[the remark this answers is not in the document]
nociale: Something like this?

#' `draws` fits the base imputation model to the observed data assuming MAR
#' according to the given multiple imputation methodology and returns the parameter
#' estimates from the fit. The model is a MMRM model that can be fit either using a ML/REML
#' approach or a Bayesian approach, according to the user's method specification. The ML/REML
#' approaches support a wide range of possible covariance structures, while the Bayesian
#' approach based on MCMC sampling supports only an unstructured covariance. In both cases
#' the covariance matrix can be assumed to be the same or different across each group.

gowerc (approved):

PR #161, opened 2021-09-17
document mcmc and add some tests

nociale:
@gowerc I documented some functions of the mcmc.R file and added some tests. There are some functions that you implemented to prepare data for the Stan fit. I did not document those as I don't know exactly what the functions do.


R/mcmc.R
```diff
@@ -114,10 +114,34 @@ fit_mcmc <- function(
...
+#' Transform an array into list of arrays where the listing
+#' is performed on a given dimension.
+#'
+#' @param a Array with number of dimensions at least 2.
+#' @param n Positive integer. Dimension of `a` to be listed.
+#'
+#' @details
+#' For example, if `a` is a 3 dimensional array and `n = 1`,
+#' `split_dim(a,n)` returns a list of 2 dimensional arrays (i.e.
+#' a list of matrices) where each element of the list is `a[i, , ]`, where
```
gowerc, line 129: Its hard to understand what this function is doing from the worded description. Perhaps just provide a small example.  For instance I did something similar in `invert_indexes()`

R/mcmc.R
```diff
@@ -136,29 +160,36 @@ split_dim <- function(a, n) {
 }
 
 
-#' Title - TODO
+#' Extract draws from a `stanfit` object
+#'
+#' @description
+#' Extract draws from a `stanfit` object, manipulate it
```
gowerc, line 166: What manipuilations does it do ?

gowerc (changes requested):

R/mcmc.R
```diff
@@ -136,29 +160,36 @@ split_dim <- function(a, n) {
...
     ##################### from array to list
-    pars$sigma <- split_dim(pars$sigma, 1)
+    pars$sigma <- split_dim(pars$sigma, 1) # list of length equal to the number of draws
 
-    if (length(dim(pars$sigma[[1]])) == 3) { # if same_cov == FALSE
-        pars$sigma <- lapply(
-            pars$sigma,
-            function(x) split_dim(x, 1)
-        )
-    } else {
```
gowerc, line 156: Looks like you've collapsed the logic for same_cov  =TRUe / FALSe into a single block, do you have unit tests that show it work in both scenarios ?

gowerc (changes requested):

R/mcmc.R
```diff
@@ -246,7 +314,11 @@ check_mcmc <- function(stan_fit, n_draws, threshold_lowESS = 0.4) {
 
 #' QR decomposition
 #'
-#' @param mat A Matrix to perform the QR decomposition on
+#' @description
+#' QR decomposition as defined in the
+#' \href{https://mc-stan.org/docs/2_27/stan-users-guide/QR-reparameterization-section.html}{Stan user's guide (section 1.2)}.
```
gowerc, line 319: Minor point, I would recommend using markdown syntax where possible.

```
#' See more about the markdown markup at the
#' [Commonmark web site](http://commonmark.org/help)
```

R/mcmc.R
nociale: Yes, the tests I have added in test-mcmc regards exactly this. Essentially the "else" was never ran since the if condition is always true.

R/mcmc.R
nociale: ok, done

R/mcmc.R
nociale: I re-phrased it better

R/mcmc.R
```diff
@@ -136,29 +187,39 @@ split_dim <- function(a, n) {
 }
 
 
-#' Title - TODO
+#' Extract draws from a `stanfit` object
+#'
+#' @description
+#' Extract draws from a `stanfit` object and convert them into lists.
+#'
+#' The function \link[rstan]{extract} returns the draws for a given parameter as an array. This function
```
gowerc, line 195: Minor but please use the markdown syntax `[rstan::extract()]`

R/mcmc.R
```diff
@@ -136,29 +187,39 @@ split_dim <- function(a, n) {
...
+#' calls \link[rstan]{extract} to extract the draws from a `stanfit` object
```
gowerc, line 196: Please use markdown syntax

R/mcmc.R
```diff
@@ -204,9 +282,20 @@ check_ESS <- function(stan_fit, n_draws, threshold = 0.4) {
...
-#' Title - TODO
+#' Diagnostics of the MCMC based on HMC-related measures.
+#'
+#' @description
+#' Check that:
+#' 1. There are no divergent iterations.
+#' 2. The Bayesian Fraction of Missing Information (BFMI) is sufficiently low.
+#' 3. The number of iterations that saturated the max treedepth is zero.
+#'
+#' Please see \link[rstan]{check_hmc_diagnostics} for details.
```
gowerc, line 293: Please use markdown syntax

R/mcmc.R
```diff
@@ -1,17 +1,52 @@
...
+#' @description
+#' `fit_mcmc()` fits the base imputation model using a Bayesian approach. This is done through a MCMC method that is implemented in `stan`
+#' and is run by using the function [rstan::sampling()].
+#' The function returns the draws from the posterior distribution of the model parameters
+#' and the `stanfit` object. Additionally it performs multiple diagnostics checks of the chain
+#' and returns warnings in case of any detected issues.
+#'
+#' @param designmat The design matrix of the fixed effects.
+#' @param outcome The response variable. Must be numeric.
+#' @param group The group variable.
```
gowerc, line 13: Maybe worth stating what types these need to be,  I.e. I think group and visit need to be factor ?

R/mcmc.R
```diff
@@ -1,17 +1,52 @@
...
+#' 1. Fit MMRM using a REML approach.
+#' 2. Prepare the input data for the MCMC fit as described in the `data{}` block of the Stan file. See [prepare_stan_data()] for details.
+#' 3. Run the MCMC according the input arguments and using as starting values the REML parameter estimates
+#' estimated at point 1.
+#' 4. Performs diagnostics checks of the MCMC. See [check_mcmc()] for details.
+#' 5. Extract the draws from the model fit.
+#'
+#' The chains perform `n_imputations` draws by keeping one every `burn_between` iterations. Additionally
+#' the first `burn_in` iterations are discarded. The total number of iterations will then be `burn_in + burn_between*n_imputations`.
+#' The `burn_in` aims to perform draws from the stationary distribution of the Markov Chain.
```
gowerc, line 42: Maybe reword this slightly?

"the purpose of `burn_in` is to ensure that the samples that are drawn are drawn from the stationary ..." ?

maybe unnecessary though

R/mcmc.R
```diff
@@ -1,17 +1,52 @@
...
+#' The `burn_between` aims to keep the draws uncorrelated each other.
```
gowerc, line 43: missing a "from"

gowerc:
@nociale  I pushed a commit adding documentation for the remaining mcmc functions that I wrote, I hope that it is clear enough ...

PR #166, opened 2021-09-20
added more documentation

gowerc:
Removed multiple TODO's in the documentation


R/impute.R
```diff
@@ -4,53 +4,59 @@
...
-#' those defined in `vars` argument as specified in the call to `draws`.
-#' You can only update the imputation method, that is to say that specifying
-#' a visit column has no effect here. If you wish to change the ICE visit you will need to re-run `draws`.
-#' A key limitation is that you can only switch between a MAR and a non-MAR method (or vice versa)
-#' if that subject has no non-missing post ICE data.
-#' As an example, if a subject had their ICE on "Visit 2" but still had values in the dataset for
-#' "Visit 3" then the function will throw and error if you try to switch them from MAR to a non-MAR method or
-#' from a non-MAR method to MAR.
+#' `impute()` works by taking the model parameter estimates, as generated by [draws()], and using
+#' them to calculate the conditional distribution for the outcome for each subject separately.
```
nociale, line 29: "conditional" -> "marginal"

R/impute.R
```diff
@@ -4,53 +4,59 @@
...
+#' These distributions are then augmented based upon the; intercurrent event (ICE) data,
```
nociale, line 30: "These distributions are then augmented based upon the intercurrent event (ICE)..."

R/impute.R
```diff
@@ -4,53 +4,59 @@
...
+#' the chosen imputation strategy, and the chosen reference group. Imputed values are then drawn
+#' from the augmented imputation distribution with 1 dataset being generated per set of parameter
```
nociale, line 32: "Imputed values are then drawn from the augmented imputation distribution after conditioning on the observed outcome values with 1 dataset being...."

R/impute.R
```diff
@@ -4,53 +4,59 @@
...
+#' estimates provided by [draws()].
+#'
+#' The exact manor in how missing values are imputed from this augmented distribution depends
```
nociale, line 35: is "manor" correct in english?

R/impute.R
```diff
@@ -4,53 +4,59 @@
...
+#' on the method object that was provided to [draws()], in particular:
+#'
+#' - Bayes & Approximate Bayes: each imputed dataset consists of 1 row per subject & visit
+#' from the original dataset with missing values imputed by taking a single random sample
+#' from the augmented outcome distribution.
+#'
+#' - Conditional Mean: each imputed dataset consists of 1 row per subject & visit from the
```
nociale, line 42: Maybe we should mention here that also the original dataset is imputed at this stage?

R/impute.R
```diff
@@ -4,53 +4,59 @@
...
+#' one for the subject ID and another for the imputation strategy, whose names are the same as
+#' those defined in `vars` argument as specified in the call to [draws()]. Please note that this
+#' argument only allows you to update the imputation strategy, that is you cannot update the ICE
+#' visit.
+#' If you do need to change the ICE visit you will need to re-run [draws()]. A key limitation
+#' of this functionality is
+#' that you can only switch between a MAR and a non-MAR strategy (or vice versa) if that subject
+#' has no observed post ICE data. As an example, if a subject had their ICE on "Visit 2"
+#' but has observed/known values for "Visit 3" then the function will throw an error
+#' if you try to switch them from MAR to a non-MAR strategy.
```
nociale, line 59: ..from non-MAR to MAR we actually throw only a warning?

nociale (changes requested):

R/impute.R
gowerc: But we calculate the conditional distribution for each subject based on their covariates ? I.e. we take the model and fit it using the patients baseline characteristics ? Thats a conditional distribution no ?

R/impute.R
gowerc: should be "manner" will update

R/impute.R
nociale: Yes, you're right, the naming is just a bit confusing since we then condition also on the observed dat after. How we usually tell the story, is that we first build the "marginal imputation distribution" for each patient, and then we build the conditional distribution conditioning on observed outcome values. But we can also keep this naming you are having, just mentioning after that we additionally condition on the observed data.

R/impute.R
gowerc: It was supposed to be a list (though I think I should have used `:` instead of `;`  Updated

R/impute.R
gowerc: Not quite sure what you mean ?

nociale: I mean, we don't just impute each bootstrap/jackknife sample but also the full original dataset. Maybe it is worthed to mention this here.

R/impute.R
gowerc: Perhaps I should be more explicit with my phrasing then perhaps:

```
#' `impute()` works by taking the model parameter estimates, as generated by [draws()], and using
#' them to calculate the distribution of the outcome variable separately for each subject by conditioning on their covariates.
```

R/impute.R
gowerc: Yer I wasn't sure about this. I had originally included a line but then removed it as I was wondering if we were duplicating text from both `draws()` and `methods()`.  If you think it still adds value I am happy to add it back in

R/impute.R
nociale: sounds good!

nociale (approved):

R/impute.R
gowerc: done

R/impute.R
gowerc: done

PR #173, opened 2021-09-22
Changes do docu of impute.R

wolbersm:
I also don't quite understand what the following means:
#' I.e. if a subject is used in samples  1,3,5,9 then these are all performed at once. 
What does this mean and how does this work? 
Wouldn't different samples typically have different imputation parameters?


R/impute.R
```diff
@@ -12,8 +12,8 @@
 #'
 #' @param draws A `draws` object created by [draws()].
 #'
-#' @param references a named vector. Identifies the references to be used when generating
-#' the imputed values. Should be of the form `c("Group1" = "Reference1", "Group2" = "Reference2")`.
+#' @param references a named vector. Identifies the references to be used when for reference-based
```
nociale, line 15: Maybe we can delete "when" from this sentence?

R/impute.R
```diff
@@ -25,42 +25,46 @@
...
-#' from the augmented imputation distribution after conditioning on the observed outcome values
-#' with 1 dataset being generated per set of parameter estimates provided by [draws()].
-#'
-#' The exact manner in how missing values are imputed from this augmented distribution depends
+#' `impute()` uses the imputation model parameter estimates, as generated by [draws()], to
+#' first calculate the marginal (multivariate normal) distribution of a subjects's longitudinal outcome variable 
+#' depending on their covariate values.
+#' For subjects with intercurrent events (ICEs) handled using non-MAR methods, this marginal distribution
+#' is then updated depending on the time of the first visit affected by the ICE, 
+#' the chosen imputation strategy; and the chosen reference group.
```
nociale, line 33: without ";"

R/impute.R
```diff
@@ -330,8 +334,8 @@ invert_indexes <- function(x) {
 #'
 #' Note that this function performs all of the required imputations for a subject at the
 #' same time. I.e. if a subject is used in samples  1,3,5,9 then these are all performed at
-#' once. The function supports patients belonging to the same sample multiple times
-#' i.e. 1,1,2,3,5,5.
+#' once. The function supports patients belonging to the same sample multiple times,
+#' i.e. 1,1,2,3,5,5, as will typical occur for bootstrapped datasets.
```
nociale, line 338: typical -> typically

gowerc:
@wolbersm 

> I also don't quite understand what the following means:
#' I.e. if a subject is used in samples 1,3,5,9 then these are all performed at once.
What does this mean and how does this work?
Wouldn't different samples typically have different imputation parameters?


Indeed they do use different imputation parameters.  The way this works is that we build a design matrix for the subject for their covariate values and then also build a beta matrix for each sample they appear in

i.e. say that
for 
- sample 1:  beta  =   1,2,3
- sample 2:  beta =  4, 5,6 
- sample 3: beta = 7,8,9

So if the patient appeared in  sample  1  twice and sample 3  once  then their beta matrix would be:

1 2 3
1 2 3
7 8 9

The main purpose of all this is that looking up the patients covariates and expanding them to the design matrix is expensive so we only do it once and then multiply them by all the different beta coefficients in one go.  We then have functions to re-group the results by sample instead of by subject.

Hope this makes sense... (apologies if the internal documentation isn't clear enough on this)

R/impute.R
wolbersm: Done

R/impute.R
wolbersm: Done

R/impute.R
wolbersm: Done

wolbersm:
> @wolbersm
> 
> > I also don't quite understand what the following means:
> > #' I.e. if a subject is used in samples 1,3,5,9 then these are all performed at once.
> > What does this mean and how does this work?
> > Wouldn't different samples typically have different imputation parameters?
> 
> Indeed they do use different imputation parameters. The way this works is that we build a design matrix for the subject for their covariate values and then also build a beta matrix for each sample they appear in
> 
> i.e. say that
> for
> 
> * sample 1:  beta  =   1,2,3
> * sample 2:  beta =  4, 5,6
> * sample 3: beta = 7,8,9
> 
> So if the patient appeared in sample 1 twice and sample 3 once then their beta matrix would be:
> 
> 1 2 3
> 1 2 3
> 7 8 9
> 
> The main purpose of all this is that looking up the patients covariates and expanding them to the design matrix is expensive so we only do it once and then multiply them by all the different beta coefficients in one go. We then have functions to re-group the results by sample instead of by subject.
> 
> Hope this makes sense... (apologies if the internal documentation isn't clear enough on this)

Thanks a lot, makes sense now! I have tried to clarify the wording in my branch.

gowerc (approved):

PR #175, opened 2021-09-22
Update pool docu

wolbersm:
Please also re-check that one- and two-sided p-values are correct for both percentile and normal/t-distribution based methods.

### DIFF
--- a/R/pool.R
+++ b/R/pool.R
@@ -1,6 +1,6 @@
 
 
-#' Pool analysis results across imputed datasets
+#' Pool analysis results
 #'
 #' @param results an analysis object created by [analyse()].
 #'
--- a/R/pool.R
+++ b/R/pool.R
@@ -1,6 +1,6 @@
 
 
-#' Pool analysis results
+#' Pool analysis results obtained from the imputed datasets
 #'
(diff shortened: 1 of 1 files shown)


gowerc (approved):

R/pool.R
```diff
@@ -1,6 +1,6 @@
 
 
-#' Pool analysis results
+#' Pool analysis results across imputed datasets
```
gowerc, line 3: @nociale ,  I am happy with all the changes here. This one here was the only one I wasn't sure about as we aren't using the imputed datasets at the stages but the results from the imputed datasets ?  I am not sure if I am just being too pedantic though, what do you think ?

nociale: Yes I understand what you mean. We actually don't use the imputed datasets here, but only the analysis results. I don't have a strong preference, maybe if we want to be extremely precise we could say "Pool analysis results obtained from the imputed datasets" or similar? @wolbersm What do you think?

nociale (approved):

R/pool.R
wolbersm: Thanks! I agree and like "Pool analysis results obtained from the imputed datasets". Have edited this branch accordingly.

PR #176, opened 2021-09-22
Docu for ancova

wolbersm:
Minimal cosmetic edits only.


R/ancova.R
```diff
@@ -2,14 +2,15 @@
 #' Analysis of Covariance
 #'
 #' Performs an analysis of covariance between two groups returning the estimated
-#' "treatment effect" and least squared means estimates of each group.
+#' "treatment effect" (i.e. the contrast between the two treatment groups) and
```
gowerc, line 5: I think maybe we should remove the "the"  i.e.   "the contrast between two treatment groups"

R/ancova.R
```diff
@@ -2,14 +2,15 @@
...
+#' "treatment effect" (i.e. the contrast between the two treatment groups) and 
+#' the least square means estimates in each group.
 #'
 #' @param data A data.frame containing the data to be used in the model
 #' @param vars A `vars` object as generated by [set_vars()]. Only the `group`,
 #' `visit`, `outcome` and `covariates` elements are required. See details.
 #' @param visits an optional character vector specifying which visits to
-#' fit the ancova model at. If `NULL` all available
-#' visits (as determined by `unique(data[[vars$visit]])`) will be used.
+#' fit the ancova model at. If `NULL`, a separata ancova model will be fit to
```
gowerc, line 12: Should that be a "for the"  or a "to the" i.e.  

" a separata ancova model will be fit to outcomes from each visits" ->
" a separata ancova model will be fit for the outcomes from each visits

nociale: Also, "separata" -> "separate"

Issue #159, opened 2021-09-16
Clarify baseline


gowerc:
We need to add some remarks in the documentation somewhere that clarify how "baseline" is (or well is not) used in the package and that users need to manually remove baseline visits if they don't want them to be apart of the model.

nociale:
I think users should be aware by themselves on what they are using as response variable :). They could also have a look at the imputed datasets to verify this. However we can write it somewhere, I am not against it.

PR #196, opened 2021-09-24
Update draws.R

gowerc:
Small paragraph to address #159

### DIFF
--- a/R/draws.R
+++ b/R/draws.R
@@ -54,10 +54,10 @@
 #' Conditional mean imputation methods are discussed in Wolbers et al (2021).
 #'
 #' The argument `data` contains the longitudinal data. It must have at least the following variables:
-#' - `subjid`: a factor containing the subject ids variable.
-#' - `visit`: a factor containing the visit variable.
-#' - `group`: a factor containing the group variable.
-#' - `outcome`: the outcome variable. Must be numeric.
+#' - `subjid`: a factor vector containing the subject ids.
+#' - `visit`: a factor vector containing the visit the the outcome was observed on.
+#' - `group`: a factor vector containing the group that the subject belongs to.
+#' - `outcome`: a numeric vector containing the outcome that was (or wasn't) observed.
 #' Additional baseline or time-varying covariates must be included in `data`.
 #'
 #' `data` must have one row per visit per subject. This means that incomplete
@@ -67,6 +67,10 @@
 #' Last Observation Carried Forward (LOCF) imputation to impute the covariates values. 
 #' Note that LOCF is generally not a principled imputation method and should only be used when appropriate 
 #' for the specific covariate.
+#'
+#' Please note that there is no special provisioning for "baseline". If you do not want baseline
+#' observations to be included in the model then these should be removed in advance from `data`. 
+#' The imputation model(s) created by `draws()` will fit the distribution to all levels of `data["visit"]`.
 #'
 #' The argument `data_ice` contains information about the occurrence of ICEs. It is a
 #' `data.frame` with 3 columns:
--- a/R/draws.R
+++ b/R/draws.R
@@ -55,7 +55,7 @@
 #'
 #' The argument `data` contains the longitudinal data. It must have at least the following variables:
 #' - `subjid`: a factor vector containing the subject ids.
-#' - `visit`: a factor vector containing the visit the the outcome was observed on.
+#' - `visit`: a factor vector containing the visit the outcome was observed on.
 #' - `group`: a factor vector containing the group that the subject belongs to.
 #' - `outcome`: a numeric vector containing the outcome that was (or wasn't) observed.
 #' Additional baseline or time-varying covariates must be included in `data`.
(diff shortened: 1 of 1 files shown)


R/draws.R
```diff
@@ -54,10 +54,10 @@
 #' Conditional mean imputation methods are discussed in Wolbers et al (2021).
 #'
 #' The argument `data` contains the longitudinal data. It must have at least the following variables:
-#' - `subjid`: a factor containing the subject ids variable.
-#' - `visit`: a factor containing the visit variable.
-#' - `group`: a factor containing the group variable.
-#' - `outcome`: the outcome variable. Must be numeric.
+#' - `subjid`: a factor vector containing the subject ids.
+#' - `visit`: a factor vector containing the visit the the outcome was observed on.
```
nociale, line 58:
```suggestion
#' - `visit`: a factor vector containing the visit the outcome was observed on.
```

R/draws.R
```diff
@@ -54,10 +54,10 @@
...
+#' - `group`: a factor vector containing the group that the subject belongs to.
+#' - `outcome`: a numeric vector containing the outcome that was (or wasn't) observed.
```
nociale, line 60:
```suggestion
#' - `outcome`: a numeric vector containing the outcome variable. It might contain missing values.
```

R/draws.R
```diff
@@ -68,6 +68,10 @@
 #' Note that LOCF is generally not a principled imputation method and should only be used when appropriate 
 #' for the specific covariate.
 #'
+#' Please note that there is no special provisioning for "baseline". If you do not want baseline
```
nociale, line 71:
```suggestion
#' Please note that there is no special provisioning for the baseline outcome values. If you do not want baseline
```

R/draws.R
```diff
@@ -68,6 +68,10 @@
...
+#' observations to be included in the model then these should be removed in advance from `data`. 
+#' The imputation model(s) created by `draws()` will fit the distribution to all levels of `data["visit"]`.
```
nociale, line 73: The imputation model(s) created by `draws()` will fit the distribution to all levels of `data["visit"]`. I think the above is clear enough and we can remove this sentence?

R/draws.R
```diff
@@ -68,6 +68,10 @@
 #' Note that LOCF is generally not a principled imputation method and should only be used when appropriate 
 #' for the specific covariate.
 #'
+#' Please note that there is no special provisioning for "baseline". If you do not want baseline
+#' observations to be included in the model then these should be removed in advance from `data`.
```
nociale, line 72:
```suggestion
#' observations to be included in the model then these should be removed in advance from the outcome variable in `data`. 
```

nociale: @gowerc I am not 100% sure we need to mention this, but I am fine with it. 

Just "remove them from `data`" might be confusing since one might want to include the baseline outcome as covariate in the model, so in this case the user should not remove it from `data`.

R/draws.R
```diff
@@ -54,10 +54,10 @@
...
 #' The argument `data` contains the longitudinal data. It must have at least the following variables:
-#' - `subjid`: a factor containing the subject ids variable.
-#' - `visit`: a factor containing the visit variable.
-#' - `group`: a factor containing the group variable.
-#' - `outcome`: the outcome variable. Must be numeric.
+#' - `subjid`: a factor vector containing the subject ids.
+#' - `visit`: a factor vector containing the visit the outcome was observed on.
+#' - `group`: a factor vector containing the group that the subject belongs to.
+#' - `outcome`: a numeric vector containing the outcome variable. It might contain missing values.
 #' Additional baseline or time-varying covariates must be included in `data`.
```
gowerc, line 61:
```suggestion
#' Additional baseline or time-varying covariates must be in `data`.
```

R/draws.R
```diff
@@ -68,6 +68,9 @@
 #' Note that LOCF is generally not a principled imputation method and should only be used when appropriate 
 #' for the specific covariate.
 #'
+#' Please note that there is no special provisioning for the baseline outcome values. If you do not want baseline
+#' observations to be included in the model then these should be removed in advance from the outcome variable in `data`.
```
nociale, line 72:
```suggestion
#' observations to be included in the model as part of the response variable then these should be removed in advance from the outcome variable in `data`. If the user wants to include the baseline outcome as covariate in the model, then the baseline outcome should be included as a separate column of `data`.
```

R/draws.R
```diff
@@ -68,6 +68,9 @@
 #' Note that LOCF is generally not a principled imputation method and should only be used when appropriate 
 #' for the specific covariate.
 #'
+#' Please note that there is no special provisioning for the baseline outcome values. If you do not want baseline
+#' observations to be included in the model as part of the response variable then these should be removed in advance from the outcome variable in `data`. At the same time if the you want to include the baseline outcome as covariate in the model, then the baseline outcome should be included as a separate column of `data` (as any other covariate).
```
nociale, line 72:
```suggestion
#' observations to be included in the model as part of the response variable then these should be removed in advance from the outcome variable in `data`. At the same time if you want to include the baseline outcome as covariate in the model, then this should be included as a separate column of `data` (as any other covariate).
```

nociale (approved):

PR #214, opened 2021-09-29
add antidepressant trial analysis

nociale:
Have added code related to the antidepressant trial analysis. Also the results from the SAS 5 macros are included


gowerc:
I can't see the R results, how do the 2 compare ?

gowerc (approved):

PR #220, opened 2021-10-04
pval = 1 when > 1 in bootstrap percentile

nociale:
The formula to get the two-sided pvalue in percentile bootstrap is the following: the formula for the pvalue is: 

```
2*min( (#theta <0 + 1)/(B+1), (#theta >0 + 1)/(B+1)  )
``` 

The pvalue has its upper limit when:

```
#theta <0 = #theta >0 = B/2
```

In this case the pvalue is 

```
2*(B/2 + 1)/(B+1) = .. = (B+2)/(B+1)
``` 

which is above 1. Of course, for large `B` then this is kind of negligible. We modify the pvalue being equal to `pval <- min(pval, 1)`.

### DIFF
--- a/R/pool.R
+++ b/R/pool.R
@@ -374,7 +374,7 @@ pool_bootstrap_percentile <- function(est, conf.level, alternative) {
         est = est_orig,  # First estimate should be original dataset
         ci = ci,
         se = NA,
-        pvalue = min(pvals[index]) * length(pvals[index])
+        pvalue = min(min(pvals[index]) * length(pvals[index]), 1)
     )
     return(ret)
 }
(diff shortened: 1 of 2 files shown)


gowerc:
Is this not a flaw in the algorithm ? This feels a bit like a duct tape solution to me; should we not use a better algorithm (if there is one) ?

gowerc:
Also please add unit tests to show this is working as expected :)

nociale:
@gowerc By different algorithm you mean a different formula? I am not sure whether we could just use `2*min( #theta <0/B, #theta >0/B  )`. In this case, the estimated p-value will be 1 anyway in the particular case mentioned above.

gowerc:
Yer sorry by algorithm I mean formula. Where did this original formula even come from ?  Did @wolbersm  get it from a paper ?

PR #221, opened 2021-10-04
fixed bug associated with not setting an initial ICE

gowerc:


R/longData.R
```diff
@@ -51,9 +51,9 @@ longDataConstructor <- R6::R6Class(
         strata = NULL,
 
 
-        #' @field visit_ice A list indexed by subject storing the visit level which
+        #' @field ice_visit_index A list indexed by subject storing the visit index number which
```
nociale, line 54:
```suggestion
        #' @field ice_visit_index A list indexed by subject storing the visit index number of
```

nociale, line 54:
```suggestion
        #' @field ice_visit_index A list indexed by subject storing the index number of the first visit affected by the ICE. It is equal to the number of visits plus 1 in case of no ICE.
```

R/longData.R
```diff
@@ -51,9 +51,9 @@ longDataConstructor <- R6::R6Class(
...
         #' the patient had their ICE on
```
nociale, line 55:
```suggestion
        #'
```

R/draws.R
```diff
@@ -100,7 +104,13 @@
 #' (performed with the function ([impute()]). To summarize, **at this stage only pre-ICE data
 #' and post-ICE data that is after ICEs for which MAR imputation is specified are used**.
 #'
-#' The vars argument is a named list that specifies the names of key variables within
+#' If the `data_ice` argument is omitted, or if a subject doesn't have a record within it, then it is
+#' assumed that a hypothetical ICE occurred after the last visit for all impacted subjects and as such all missing data
```
nociale, line 108:
```suggestion
#' assumed that no ICEs occurred and as such all missing data
```

R/draws.R
```diff
@@ -100,7 +104,13 @@
...
+#' will be imputed under MAR. Please note that the ICE visit cannot be changed by the `update_strategy` argument
+#' in [impute()] and as such the subjects missing data will still be imputed under MAR even if you change the
+#' subjects imputation strategy.
```
nociale, line 111: Is this comment needed? E.g.  when changing from MAR to non-MAR we should throw an error.

gowerc: It won't throw an error as there is no Post-ICE data

nociale: ok, so in this case if the user changes the imputation strategy to "JR" in `impute()`, the function will impute again under MAR? It makes sense to me, since the ICE did not occurr and so setting JR is just meaningless. Should we throw a warning, e.g. "you are setting a JR strategy for a patient that did not experienced any ICE. The imputation will be performed under MAR" (also for CRAN release if now there is no time).

gowerc: My gut feeling is that would be overkill. At least we have it clear in the documentation that not setting the ice visit locks it at visit + 1  and also that pre-ice visits are always imputed with MAR.

I'm also feeling that if someone was being lazy and just did a blanket "impute everything with JR" it would be annoying to then have to go through and individually remove all the patients that didn't have an ICE.

nociale: agreed!

nociale (approved):

Issue #238, opened 2021-10-22
Add conservative adjustment if get_jackknife_draws fails for a leave-one-subject-out dataset.


wolbersm:
For method_condmean(type="jackknife"), it may occur that the estimator fails for only one or very few leave-one-subject-out datasets. Currently, the method fails altogether in this setting and this is ok for 99.9% of the cases, as this should be extremely rare. However, it still seems reasonable to build in a relatively conservative handling of this "worst case scenario" to avoid complete failure.

My proposal is to:
a) Add a tolerance threshold which by default is 0 but allows up to x% jackknife failures.
b) If the MMRM/estimator fails on a leave-one-subject-out datasets, treat it as follows: 
    - the corresponding draws, imputed dataset and estimators are NA/NULL
    - for the jackknife SE calculation, the estimators for all failed leave-one-out dataset are conservatively set  as equal to the observed estimator amongst all leave-one-out datasets which didn't fail and which had the largest absolute distance to the estimator from the original dataset.

nociale:
@wolbersm I have a theoretical question. Let's denote by `theta` the estimate of the parameter of interest using all subjects and `theta_m` the average of all estimates that come from the leave-one-out samples. `theta` and `theta_m` differs in general.

The jackknife variance estimator is `(N-1)/N * sum( (theta_i - **theta_m**)^2 )`. If we have failures we don't have some estimates of some leave-one-out samples. That is, we do not have `theta_m` but we do have `theta`. The jackknife variance estimator in this case is:

1. Compute `theta_i - theta` for all non failed estimates.
2. Set all failed estimates equal to `theta_max = max(theta_i - theta)`. 
3. Now we have estimates for all leave-one-out samples. We can apply the formula `(N-1)/N * sum( (theta_i - mean(theta_i))^2 )` (this time the index `i` includes also the "estimates" derived at point 2).

Questions: 
- Is this procedure what you mean?
- This "conservative jackknife variance estimator" at point 2 maximizes the distance from `theta` and not from `mean(theta_i) i = set of non-failed estimates`. Is this the correct way to go? Does this maximization provides the same `theta_max` of `max(theta_i - theta_m), i = set of non-failed estimates` if I would have observed `theta_m`?

Thanks!

wolbersm:
@nociale 
I think in step 1, one would have to set the estimates in the failed datasets to theta_max=theta+max(abs(theta_i-theta)) instead.  Does this make sense?

nociale:
[...]

PR #243, opened 2021-11-01
improve error messages for mmrm failure in jackknife or bayes methods

### LINKED ISSUE #237: Nicer error message when MMRM fails in Bayes
When `method <- method_bayes()`, if MMRM fails, th error message that appears is:

`Error in stan_data$R %*% mmrm_initial$beta : requires numeric/complex matrix/vector arguments`

Indeed, `mmrm_initial` is a list of one element: `mmrm_initial$failed` which is `TRUE`.

A nicer error message, like the one we use for analogous case for the `condmean(type = "bootstrap")` methods, would be better:

`Fitting MMRM to original dataset failed`

Also `condmean(type = "jackknife")` does not distinguish between error in MMRM fit in the original dataset or in a leave-one-out sample: the error message is the same and is the following: `Jackknife sample failed`

nociale:
Closes #237.


nociale:
Last commit also addresses #238

R/draws.R
```diff
@@ -329,7 +327,9 @@ get_jackknife_draws <- function(longdata, method) {
             optimizer = optimizer
         )
         if (sample$failed) {
-            stop("Jackknife sample failed")
+            stop(sprintf("Fitting MMRM to jackknife sample without subject with ID %s failed",
```
gowerc, line 330:
```suggestion
            stop(sprintf("MMRM failed to fit to the data after removing subject %s",
```

nociale:
@gowerc Can we merge?

R/draws.R
```diff
@@ -325,7 +327,9 @@ get_jackknife_draws <- function(longdata, method) {
             optimizer = optimizer
         )
         if (sample$failed) {
-            stop("Jackknife sample failed")
+            stop(sprintf("MMRM failed to fit to the data after removing subject %s",
+                         ids[i])
```
gowerc, line 331:
```suggestion
            stop(
                sprintf("MMRM failed to fit to the data after removing subject %s", ids[i])
```

R/mcmc.R
```diff
@@ -62,14 +54,16 @@ fit_mcmc <- function(
     group,
     subjid,
     visit,
-    same_cov,
-    n_imputations,
-    burn_in,
-    burn_between,
-    verbose = TRUE,
-    seed = NULL
+    method
```
gowerc, line 57: I'm suprised this didn't require any changes to our unit tests. Do we not have anything for fit_mcmc ?

R/mcmc.R
```diff
@@ -83,6 +77,10 @@ fit_mcmc <- function(
         optimizer = c("L-BFGS-B", "BFGS")
     )
 
+    if (mmrm_initial$failed) {
+        stop("Fitting MMRM to original dataset failed")
+    }
+
```
gowerc, line 83: What happened if it failed before this ?  Would it just carry on processing ?

R/mcmc.R
nociale: I just checked that we have some tests for fit_mcmc on nightly mode.. I will make changes accordingly

R/mcmc.R
nociale: It was giving a less interpretable error when operating with the regression coefficients which where NA probably: 

`Error in stan_data$R %*% mmrm_initial$beta : requires numeric/complex matrix/vector arguments`

R/mcmc.R
nociale: ok, just pushed the changes

R/mcmc.R
gowerc: ouch , yer good to have this here now !

Could you include a test for a model which can't converge and show that this catches it though please.  I know its trivial but its good to have encase we edit the code in the future and break this for some reason

nociale: Ok, I will do it!

nociale: Done!

gowerc (approved):

PR #247, opened 2021-11-04
Rubin na

### LINKED ISSUE #226: Allow Rubin's rules to work in case of `se = NA`?
Sometimes one might want to get point estimates without SE estimation. 

For example, for a responder analysis the analysis model might be a logistic regression for a dichotomized endpoint. One might additionally want the estimated proportion of responders in the two groups and not just the adjusted difference between them in log-odds scale, which is a regression coefficient in the logistic regression used to study significance but not easily interpretable.

This could be done by simply allowing the (user-defined) analysis model to return `se = NA` for the proportions parameters. The pooling function should simply in this case return `NA` for SE, CIs and p-value but still return the pooled point estimate.

nociale:
Closes #226 .

### DIFF
--- a/R/analyse.R
+++ b/R/analyse.R
@@ -464,7 +464,7 @@ validate_analyse_pars <- function(results, pars) {
                 all(!is.na(vapply(results_unnested, function(x) x[[par]], numeric(1)))),
                 msg = sprintf("Parameter `%s` contains missing values", par)
             )
-        } else if(par == "se") {
+        } else {
             assert_that(
                 all(!is.na(vapply(results_unnested, function(x) x[[par]], numeric(1)))) ||
                 all(is.na(vapply(results_unnested, function(x) x[[par]], numeric(1)))) ,
--- a/R/analyse.R
+++ b/R/analyse.R
@@ -459,10 +459,16 @@ validate_analyse_pars <- function(results, pars) {
     )
 
     for (par in pars) {
-        if (par != "df") {
+        if (!par %in% c("df", "se")) {
             assert_that(
                 all(!is.na(vapply(results_unnested, function(x) x[[par]], numeric(1)))),
                 msg = sprintf("Parameter `%s` contains missing values", par)
+            )
+        } else if(par == "se") {
+            assert_that(
+                all(!is.na(vapply(results_unnested, function(x) x[[par]], numeric(1)))) ||
+                all(is.na(vapply(results_unnested, function(x) x[[par]], numeric(1)))) ,
+                msg = sprintf("Parameter `%s` contains both missing and observed values", par)
             )
         }
     }
--- a/R/pool.R
+++ b/R/pool.R
@@ -107,7 +107,7 @@ pool <- function(
 #' either `"rubin"`, `"jackknife"` or "`bootstrap"`.
 get_pool_components <- function(x) {
     switch(x,
-           "rubin" = c("est", "df", "se"),
+           "rubin" = c("est", "se", "df"),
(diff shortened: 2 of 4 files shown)


nociale:
@gowerc Sorry Craig I don't get what is the cause of the failure of the check..

gowerc:
I think its a issue with one of the current PROD cran packages. I need to investigate it a bit more, either way its just a problem with the master branch if you just make the PR to develop you shouldn't see it hopefully

tests/testthat/test-analyse.R
```diff
@@ -48,6 +48,26 @@ test_that("basic constructions of `analysis` work as expected",{
     )
     expect_true(validate(x))
 
+
+    x <- as_analysis(
+        results = list(
+            list(p1 = list("est" = 1, "df" = 4, "se" = NA)),
+            list(p1 = list("est" = 2, "df" = 3, "se" = NA))
+        ),
+        method = method_bayes(n_samples = 2)
```
gowerc, line 57: Should this not error due to the different DF  ?

R/analyse.R
```diff
@@ -459,11 +459,17 @@ validate_analyse_pars <- function(results, pars) {
...
             assert_that(
                 all(!is.na(vapply(results_unnested, function(x) x[[par]], numeric(1)))),
                 msg = sprintf("Parameter `%s` contains missing values", par)
             )
+        } else if(par == "se") {
+            assert_that(
+                all(!is.na(vapply(results_unnested, function(x) x[[par]], numeric(1)))) ||
+                all(is.na(vapply(results_unnested, function(x) x[[par]], numeric(1)))) ,
+                msg = sprintf("Parameter `%s` contains both missing and observed values", par)
+            )
```
gowerc, line 472: Should the df also be handled by the second else if condition ? As in we allow DF to be NA but surely we don't allow for a mix of  missing and non-missing ?

tests/testthat/test-pool.R
```diff
@@ -120,9 +136,61 @@ test_that("pool", {
...
+    results_bayes <- as_analysis(
+        method = method_bayes(n_samples = 5000),
+        results =
+            lapply(
+                seq_len(5000),
+                function(x) runanalysis(sample(vals, size = n, replace = TRUE))
+            )
+    )
+    bayes <- pool(results_bayes)
+
```
gowerc, line 184: Might be worth running this with a few different options like alternative and conf.int just to make sure theres no interactions

R/pool.R
```diff
@@ -107,7 +107,7 @@ pool <- function(
 #' either `"rubin"`, `"jackknife"` or "`bootstrap"`.
 get_pool_components <- function(x) {
     switch(x,
-           "rubin" = c("est", "df", "se"),
+           "rubin" = c("est", "se", "df"),
```
gowerc, line 110: Any reason for the switch here ?

tests/testthat/test-analyse.R
nociale: I think we are not verifying that all the df are the same in `validate.rubin` of the analysis output.

R/analyse.R
nociale: Agreed!

R/pool.R
nociale: No :( , I think I changed the order by mistake

tests/testthat/test-pool.R
nociale: Ok!

gowerc (approved):

PR #249, opened 2021-11-09
pvalue percentile

### LINKED ISSUE #232: Update pval calculation for bootstrap percentiles
I spent some time thinking about how we invert the bootstrap percentile CI to derive a p-value. 

The one-sided (upper) 1-alpha bootstrap percentile CI is [q_alpha, Inf] where q_alpha is the alpha quantile [type=6] of the bootstrap estimates. 

The p-value for H_0: theta=0 vs H_A: theta>0 is thus the value alpha for which q_alpha=0 [or the largest such value if multiple q_alpha=0, or 0 if all bootstrap estimates are > 0, or 1 if all bootstrap estimates are <0]. 

Our current implementation calculates this as p("greater") = (sum(est<0)+1)/(length(est)+1)) which I took from [Efron and Tibshirani, 1994, page 214-215; adding +1 to comply with the percentile bootstrap definition of Davison and Hinkley, 1997]. 
I believe, this is exactly correct according to the explanation above if and only if exactly one bootstrap estimate is equal to 0. 
If all bootstrap estimates are !=0 [which is the typical case], I think it is conservative and never off by more than 1/(B+1). As 1/(B+1) is very small for large B, this should not be a practical concern.  
If many bootstrap estimates are exactly = 0 [which should not really occur in practice for reasonable data], then the current calculation is more problematic. In particular, in the extreme case where all bootstrap estimates are exactly = 0, then the current calculation gives p("greater")=1/(B+1) which does not make sense. 

Therefore, I have thought about how to improve this to also give correct values in such extreme cases and my current function of creating one-sided p-values is below. Unfortunately, it is more involved than I hoped...
 
The two-sided p-value would be calculated as before with truncation at 1 because it is still possible that both one-sided p-values are >0.5 in case of tied bootstrap estimates at 0. 

I would really appreciate if either of you could think this through again carefully to make sure that I haven't messed anything up and then incorporate it in the next release. Thanks!

```
pval_percentile <- function(est){
  # Determines the (not necessarily unique) quantile (type=6) of "est" which gives a value of 0
  # From this, derive the p-value corresponding to the percentile bootstrap via inversion  
  if (all(est==0)){ # non-unqiue solution
    qmin <-0
    qmax <- 1
  } else if (min(est) > 0){ # if 0 below range of est, still give value 0
    qmin <- qmax <- 0 
  } else if (max(est)< 0) { # if 0 above range of est, still give value 1
    qmin <- qmax <- 1 
  } else if (any(est==0)){ # non-unqiue solution if multiple est==0 
    quant <- rank(est,ties.method="first")/(length(est)+1) # see ?quantile "type=6" section for explanation
    qmin <- min(quant[est==0])
    qmax <- max(quant[est==0])
  } else { 
    qmin <- qmax <- 
      uniroot(function(q) quantile(est,probs=q, type=6),lower=0, upper=1)$root
  }
  c("pval_less" = 1-qmin, "pval_greater" = qmax)
}
```

nociale:
Closes #232


R/pool.R
```diff
@@ -378,6 +382,43 @@ pool_bootstrap_percentile <- function(est, conf.level, alternative) {
...
+        qmin <- min(quant[est == 0])
+        qmax <- max(quant[est == 0])
+    } else {
+        qmin <- qmax <-
+            uniroot(function(q) quantile(est, probs = q, type = 6), lower = 0, upper = 1)$root
+    }
+
+    ret <- c("pval_greater" = qmax, "pval_less" = 1 - qmin)
+    return(ret)
+}
```
gowerc, line 421:
```suggestion
    if (all(est == 0)) {
        ret <- c(0, 0)
    } else if (min(est) > 0) {
        ret <- c(0, 1)
    } else if (max(est) < 0) {
        ret <- c(1, 0)
    } else if (any(est == 0)) {
        # see ?quantile "type=6" section for explanation
        quant <- rank(est, ties.method="first")/(length(est) + 1) 
        ret <- c(
            min(quant[est == 0]),
            1 - max(quant[est == 0])
        )
    } else {
        x <- uniroot(function(q) quantile(est, probs = q, type = 6), lower = 0, upper = 1)$root
        ret <- c(x, 1-x)
    }
    names(x) <- c("pval_greater", "pval_less")
    return(ret)
}
```

R/pool.R
```diff
@@ -340,7 +340,11 @@ pool_bootstrap_percentile <- function(est, conf.level, alternative) {
     }
 
     alpha <- 1 - conf.level
-    pvals <- (c(sum(est < 0), sum(est > 0)) + 1) / (length(est) + 1)
+
+    pvals <- pval_percentile(
+        est = est
+    )
```
gowerc, line 346:
```suggestion
    pvals <- pval_percentile(est = est)
```

gowerc:
@nociale , looks like that change broke the tests :( can you look into it ?

R/pool.R
```diff
@@ -378,6 +380,48 @@ pool_bootstrap_percentile <- function(est, conf.level, alternative) {
...
+        )
+    } else {
+        x <- uniroot(
+            function(q) quantile(est, probs = q, type = 6), 
+            lower = 0, 
+            upper = 1
+        )$root
+        ret <- c(x, 1-x)
+    }
+    names(x) <- c("pval_greater", "pval_less")
```
nociale, line 422:
```suggestion
    names(ret) <- c("pval_greater", "pval_less")
```

nociale:
@gowerc I think it was at the last line of the `pval_percentile` function: `names(x)` -> `names(ret)`

nociale:
@gowerc Also, if all est == 0, the pvalues should be 1 (H0 should be "fully" accepted, sorry for the improper terminology)

gowerc:
@nociale ,  Good spot. I think code wise I am happy with this though would you mind responding to my comment in the original issue thread please ?

gowerc (approved):

gowerc:
@nociale , please can you resolve the merge conflict here?

PR #251, opened 2021-11-10
add one arm trial support

### LINKED ISSUE #144: Should we support #groups = 1  (1 arm trial) ? And if so how ?

nociale:
Closes #144


R/utilities.R
```diff
@@ -53,14 +53,25 @@ has_class <- function(x, cls) {
...
+        is_multi_groups,
+        c(
+            vars$group,
+            vars$visit,
+            vars$covariates
+        ),
+        c(
+            vars$visit,
+            vars$covariates
+        )
```
gowerc, line 72: I'm wondering if there is a cleaner way to do this.  Like for example if its a single group could we not just set `vars$group` to NULL ?  This would save having to pass this `is_multi_groups` option all over the place :(

nociale: True. In my opinion it does not change that much. The current solution I would say is conceptually cleaner since the part related to the fixed effect formula is all in the same function instead of adapting the formula outside it. However yes, it needs this additional `is_multi_groups` argument. Let me know what you prefer, I am inclined to say that it does not matter, but I am happy to change if you think it is better.

gowerc:
@nociale ,

I've merged in develop and resolved some merge conflicts. Would you be able to add some print unit tests for this, in particular I am not sure what pool will look like for this case ?

gowerc (changes requested):
Its not clear to me at the moment what the print message should be for pool. Also please can we add some unit tests in for it.

nociale:
@gowerc Sorry, what issues could we have with the pool print?

Toy example: Suppose we have a one arm trial. After a MAR imputation, our analysis just estimates the mean at the last visit. Then the print of the pooled object just prints point estimate, standard error, CIs and p-value for the mean at the last visit.

gowerc:
@nociale ,

Just to recap,  I've added in an addition check into longdata that ensures that all factor variables listed in `covariates` have at least two levels and updated your unit tests to look for this error accordingly.  That being said they now fail due to:

```
expect_error(
            draws(
                data = dat,
                data_ice = dat_ice,
                vars = vars_wrong3,
                method = method
            ),
            "`group`"
        )

> vars_wrong3
$subjid
[1] "id"

$visit
[1] "visit"

$outcome
[1] "outcome"

$group
[1] "group_wrong"

$covariates
[1] "age"   "sex"   "visit"

$strata
[1] "group_wrong"

$strategy
[1] "strategy"
```

I wasn't sure why this one was expected to fail. My assumption would have been that given group_wrong only has 1 level that our code excludes it from the formula (though it doesn't appear to be doing that either)

nociale:
Hi @gowerc 

I just looked into this error. vars_wrong3 is a wrong specification that a user might do. Basically, the "group_wrong" variable is a 2 levels variable ("A", "B") but only one is observed ("A"). This happens if you subset the dataset selecting e.g. only patients from the intervention arm, but then you forget to re-factor the group variable to have 1 level only. 

rbmi checks whether group_wrong has more than 1 level, and this is TRUE, so this variable is added in the imputation model. However the model fit fails since group_wrong has only one level observed. 

How do you think we should address this situation?

gowerc:
I would suggest that we enhance data validation check to also consider values in the vector as well as the labels.  If you want to include this I would recommend adding it here: https://github.com/insightsengineering/rbmi/blob/eaf37a7aed50cee773d07ae739652a7ba3606aac/R/validate_datalong.R#L96

R/validate_datalong.R
```diff
@@ -98,11 +98,27 @@ validate_datalong_types <- function(data, vars) {
...
     )
 
+    assert_that(
+        length(unique(data[[vars$group]])) == nlevels(data[[vars$group]]),
+        msg = sprintf(
+            "The number of levels in variable `%s` is different than the number of observed levels",
+            vars$group
+        )
+    )
+
```
nociale, line 108: @gowerc Do you like something like this to check that the number of levels of `group` are the same as the observed levels? Or would you rather check that `unique(data[[vars$group]])) == levels(data[[vars$group]]`?

gowerc: I think what you have here already is fine.

gowerc:
I think I'm happy to merge this now unless there is anything else you want to do with it ?

nociale:
@gowerc We can merge :)

PR #253, opened 2021-11-11
unique warning when passing from non-MAR to MAR

### LINKED ISSUE #223: Unique warning message when passing from non-MAR to MAR

nociale:
Closes #223


gowerc:
Just to clarify, is the issue being solved here that it would print out the warning multiple times ?

gowerc (approved):

gowerc:
Made some tweaks to the unit test code to use our pre-existing function to catch errors but otherwise this LGTM

PR #260, opened 2021-11-16
add a test on expected results for 3 arms trial

nociale:


nociale:
Just added a test which verifies that if I have 3 groups, "A", "B" and "C" where "B" and "C" are identical, then they have same imputed values and same estimated treatment effect when compared to "A".

gowerc (approved):

PR #263, opened 2021-11-19
pre-create print objects

gowerc:
@nociale ,

This PR should address all the issues with reproducibility on the print unit tests. Essentially I have pre-generated the objects to be printed and store them within the package.  Note that there are limits on this that the .rda file must apparently be < 1Mb  to be accepted on CRAN so we probs should avoid this where possible (I'm not really happy with this but I can't think of any other way around it).

Note that we need to be careful now that if we change any of the objects we need to make sure we re-run the code to recompute the objects that need to be printed as well as re-running the snapshots to ensure they are up-to-date....


nociale:
@gowerc Thanks. Can I ask you why the current print testing was not good?

gowerc:
@nociale , Because a lot of the numbers had to be EXACTLY reproducible making it quite fragile when running the tests on different machines (R does not guarantee reproducibility across different versions of R). Stan was a particularly bad example of this. My fear is that could lead to problems say if the FDA or someone tried to re-run our test suite on a different version  / OS  / CPU and then the test failed.

gowerc:
as further evidence of why this is needed the last PR restarted the nightly builds and the first thing it did was error due to the print methods :) 

https://github.com/insightsengineering/rbmi/runs/4282259417?check_suite_focus=true

nociale:
@gowerc Thanks for the explanation!

nociale (approved):

PR #267, opened 2021-12-06
only head of imputed datasets

nociale:
Only print head of imputed datatasets in quickstart vignette


gowerc:
Just to double check that you explicitly wanted to merge this into master and not develop ?

nociale:
@gowerc Yes, I would like to merge this into master because I sent Paul the pdf of the vignette which contains this "head(imputed_df)" modification already.

gowerc (approved):

PR #269, opened 2021-12-16
Simulate data

### LINKED ISSUE #225: Better simulation function
It might be useful to have a better data simulation function. In particular the following enhancements need to be discussed:

- 1 ICE (treat disc) + given prob of drop-out
- Simulation of the probability of the intercurrent event: depends on current outcome value according to model. Need to think of input arguments to adjust this.
- Simulation of post-ICE data. MAR or reference-based assumptions are probably enough

nociale:
Closes #225.

@gowerc Two things are still missing:
- The data simulation function allows to specify a custom strategy for the post-ICE data. I.e. post-ICE data might follow one of the implemented strategies (MAR, CIR, CR, JR, LMCF) or a user-defined strategy. I just did not exactly know how to validate a user-defined strategy (I guess we have some code somewhere that does it). Could you please let me know what function validates a user-defined strategy?
- The `prob_ice1` argument is one of the arguments to specify the probability of the ICE for each visit. Currently it could be set as a number or a vector of length equal to the length of `mu`. If a vector, its last element specifies the probability that the ICE happens at the last visit, i.e. the first visit affected by the ICE will be the "visit after the last visit". In other words, the last element of `prob_ice1` is useless and the result will be the same regardless of what value is chosen. So, I thought of removing it and let `prob_ice1` be of length equal to the number of visits minus 1. However, it might be confusing for the user that he cannot specify a probability of the ICE for each visit. What do you think is best?

Best regards,
Alessandro


R/simulate_data.R
```diff
@@ -0,0 +1,611 @@
...
+        prob_ice <- rep(prob_ice, n_visits)
+    }
+    prob_ice[prob_ice == 0] <- 1e-20
+    prob_ice[prob_ice == 1] <- 1-1e-15
+
+    create_model_ice <- function(n_visits, baseline_mean) {
+        model_ice <- "~ 1"
+        for(visit_num in seq.int(n_visits) - 1) {
+            model_ice <- paste0(model_ice, " + I(visit == ", visit_num,")")
+        }
```
gowerc, line 422: Apologies if I'm misunderstanding this code but it looks like you are wanting to set the last visit to be the contrast?  If thats the case then I think you can just use the following:

```
contrasts(data$visit)  <- contr.treatment(nvisit, nvisit)
```

R/simulate_data.R
```diff
@@ -0,0 +1,611 @@
...
+        model_ice <- paste0(model_ice, " + I((x - ", baseline_mean,"))")
```
gowerc, line 423: Why are you embedding the values into the formula environment instead of just putting them in the main dataframe ?

R/simulate_data.R
```diff
@@ -0,0 +1,611 @@
...
+#' affected by the drop-out following ICE1 and `0` otherwise.
+#' - `dropout2`: Binary variable that takes value `1` if the corresponding visit is affected
+#' by the uninformative study drop-out.
+#' - `ind_ice2`: Binary variable that takes value `1` if the corresponding visit is affected
+#' by ICE2.
+#' - `outcome`: Numeric variable that specifies the longitudinal outcome including ICE1, ICE2
+#' and the intermittent missing values.
+#'
+#' @export
+simulate_data <- function(pars_c, pars_t, post_ice_traj, strategies = getStrategies()) {
```
gowerc, line 222: Do you think it would be worth setting some default values for pars_c and pars_t so that users can just run the function `simulate_data()` if they just want some simple example data without having to put any thought into it ? (i.e. to just mess around)

R/simulate_data.R
```diff
@@ -0,0 +1,611 @@
...
+#' - `outcome_noICE`: Numeric variable that specifies the longitudinal outcome assuming
+#' no ICEs.
+#' - `ind_ice1`: Binary variable that takes value `1` if the corresponding visit is
+#' affected by ICE1 and `0` otherwise.
+#' - `dropout_ice1`: Binary variable that takes value `1` if the corresponding visit is
+#' affected by the drop-out following ICE1 and `0` otherwise.
+#' - `dropout2`: Binary variable that takes value `1` if the corresponding visit is affected
+#' by the uninformative study drop-out.
+#' - `ind_ice2`: Binary variable that takes value `1` if the corresponding visit is affected
+#' by ICE2.
```
gowerc, line 217: Sorry I'm confused by these two variables, shouldn't they always be the same? From the description in the `set_simul_pars` function my understanding was that ICE2's are the cause of non-informative dropout ?

R/simulate_data.R
```diff
@@ -0,0 +1,611 @@
...
+        length(pars_c$mu) == length(pars_t$mu),
+        msg = "The two groups must have the same number of visits"
+    )
+
+    # TODO: validate user-defined strategies
+    match.arg(
+        arg = post_ice_traj,
+        choices = names(strategies)
+    )
+    strategy_fun <- strategies[[which(names(strategies) == post_ice_traj)]]
```
gowerc, line 241:
```suggestion
    strategy_fun <- strategies[[post_ice_traj]]
```

Isn't this a named list? Do we need to convert to indexing as opposed to just using the name ?

R/simulate_data.R
```diff
@@ -0,0 +1,611 @@
...
+    model_ice <- create_model_ice(
+        n_visits = n_visits,
+        baseline_mean = baseline_mean
+    )
+
+    model_coef <- c(0, log(prob_ice/(1-prob_ice)), log(or_outcome_ice))
+
+    lp <- c(
+        model.matrix(model_ice, data.frame("visit" = visits, "x" = as.numeric(outcome))) %*% model_coef
+    ) # linear predictor
```
gowerc, line 436:
```suggestion
    model_coef <- c(0, log(prob_ice/(1-prob_ice)), log(or_outcome_ice))

    dat <- data.frame(
        visits = visits,
        x = outcome - baseline_mean
    )
    mod_no_intercept <- model.matrix(~ 0 + visits + x, dat)
    mod <- cbind(matrix(1, nrow = nrow(dat), ncol = 1), mod_no_intercept)
    lp <- mod %*% model_coef
```

R/simulate_data.R
```diff
@@ -0,0 +1,611 @@
...
+    probs_ice <- binomial(link = "logit")$linkinv(lp)
+    ind_ice <- rbinom(n = length(ids), size = 1, prob = probs_ice)
+    ind_ice <- unlist(
+        tapply(
+            ind_ice, factor(ids, levels = unique(ids)),
+            function(x) c(0, pmin(cumsum(x), 1))[-(length(x) + 1)]
```
gowerc, line 442:
```suggestion
            function(x) c(0,cummax(x))[-(length(x) + 1)]
```

R/simulate_data.R
```diff
@@ -0,0 +1,611 @@
...
+        )
+    )
+    names(ind_ice) <- NULL
```
gowerc, line 445:
```suggestion
    ind_ice <- unlist(
        tapply(
            ind_ice, factor(ids, levels = unique(ids)),
            function(x) c(0, cummax(x))[-(length(x) + 1)]
        ),
        use.names = FALSE
    )
```

R/simulate_data.R
nociale: Good point. You are right.

nociale: I will only keep ind_ice2

R/simulate_data.R
nociale: At the moment, the easiest is that the user sets only mean trajectories and covariance matrices (using `set_simul_pars()`), and what happens is that data are simulated without any ICE. Is that simple enough? I think every longitudinal data simulation requires at least the specification of mean and covariances.

R/simulate_data.R
```diff
@@ -0,0 +1,611 @@
...
+    first_ice1_visit <- unlist(
+        tapply(data$ind_ice1, factor(data$id, levels = unique(data$id)), function(x) {
+            subset <- rep(0, n_visits)
+            if(any(x == 1)) {
+                ind <- which(x==1)[1]
+                subset[ind] <- 1
+            }
+            return(subset)
+        }
+        ))
```
gowerc, line 323:
```suggestion
    first_ice1_visit <- unlist(
        tapply(
            data$ind_ice1, 
            factor(data$id, levels = unique(data$id)),
            function(x) {
                subset <- rep(0, n_visits)
                subset[which(x == 1)[1]] <- 1
                return(subset)
            }
        ),
        use.names = FALSE
    )
```

R/simulate_data.R
```diff
@@ -0,0 +1,611 @@
...
+#' drop-out. If `NULL` then
+#' `subset` will be set to `1` for all the values except the values corresponding to the
+#' baseline outcome, since baseline is supposed to not be affected by drop-out.
+#' Even if `subset` is specified by the user, the values corresponding to the baseline
+#' outcome are still hard-coded to be `0`.
+#'
+#' @importFrom stats rbinom
+simulate_dropout <- function(prob_dropout, ids, subset = NULL) {
+
+    if(is.null(subset)) {
```
gowerc, line 475: Why would subet ever be NULL ?

gowerc:
You have `factor(ids, levels = unique(ids))` in quite a few locations, would it make sense to just do this once at the very start of `generate_data_single` to avoid repeating it everywhere ?

R/simulate_data.R
gowerc: I'm more thinking of people just wanting to run the functions to see how it works, I think for them it would be nice to be able to just run `simulate_data()` and have some data available even if that data is meaningless. I.e. just a quick easy way to use the primary rbmi functions without having to configure anything ?

R/simulate_data.R
nociale: ok, as I mentioned you in the chat these two are indeed not the same thing. ice2 happens if dropout2 happens and the patient hadn't ice1 before. In case that ice1 happened at visit 2, dropout1 did not happen and dropout2 happened at visit 3, ice2 does not happen (since the patient has already discontinued from treatment) but there are some missing values due to dropout2 without this being an ice. 

However we might simplify this a bit and make dropout2 and ice2 be the same (according to the definition of ice2). That is, ice2 happens as it currently happens, but we avoid additional missing values after ice1 due to dropout2.
@wolbersm what do you think? Would you keep it as originally designed or would you make this little simplification?

R/simulate_data.R
nociale: Couldn't we use `simulate_test_data()` for this purpose? I think it is currently set-up to work without any argument, isn't it? I am not sure I would like to set defaults for mean and covariances here.. How important is in your opinion to have this feature in `simulate_data()`?

R/simulate_data.R
nociale: Agreed

R/simulate_data.R
nociale: I could set default to all ones...

R/simulate_data.R
wolbersm: @nociale 
Thanks a lot for this. 
I thought about this again and, to me, it would be better so simulate ICE2 (NSDRC discontinuations) slightly differently. 
My proposal is to omit argument prob_dropout altogether and have instead the following arguments for ICE2 (NSDRC discontinuations):
- drop_ice2 [constant visit-wise probability of ICE2] 
- prob_post_ice2_dropout
I.e. ICE2 would be handled essentially in the same way as for ICE1 except that it is simpler in the sense that we would not include an argument or_outcome_ice2. 

If for a subject, both ICE1 and ICE2 are simulated to occur, then only the earlier of them counts. In case both ICEs are simulated to occur at the same time, I guess we can decide that we count is as ICE1.

Apologies for late suggestion! If this is complicated to change or you don't like the solutation, please also let me know.

nociale:
From @gowerc comments:

- [ ] factor(ids, levels = unique(ids)) is done in 4 different locations. Perhaps its worth just applying this once at the start of the data generation.
- [ ]  After ice1 the control group shouldn't have their outcome updated (even if from the same distribution).

gowerc (approved):
I'm happy with the current code. I guess make sure you are happy with Marcels comments or resolve them before accepting the merge.

wolbersm:
Sorry @nocia, one more comment: I saw that the function has an argument `post_ice_traj`. 
I think it would be best to rename this to `post_ice1_traj` for additional clarity that this refer to ICE1 only.

gowerc:
@nociale 

regarding

> The prob_ice1 argument is one of the arguments to specify the probability of the ICE for each visit. Currently it could be set as a number or a vector of length equal to the length of mu. If a vector, its last element specifies the probability that the ICE happens at the last visit, i.e. the first visit affected by the ICE will be the "visit after the last visit". In other words, the last element of prob_ice1 is useless and the result will be the same regardless of what value is chosen. So, I thought of removing it and let prob_ice1 be of length equal to the number of visits minus 1. However, it might be confusing for the user that he cannot specify a probability of the ICE for each visit. What do you think is best?

Yer I was wondering about this and found it confusing myself.  I think it is much more intuitive for the probability to apply to the visit in question rather than the subsequent visit  i.e.  if I specified  `c(0, 0.5, 0.7)` it would make more sense to me if Visit 0  had  0% ICE's,    visit 1 had 50% ICE's  and visit 2 had 70% ICES.   I would then just put in the documentation that the first digit is ignored as baseline will always have 0% ICE's.  I really don't like the last digit being ignored.

gowerc:
@nociale 

Regarding

> The data simulation function allows to specify a custom strategy for the post-ICE data. I.e. post-ICE data might follow one of the implemented strategies (MAR, CIR, CR, JR, LMCF) or a user-defined strategy. I just did not exactly know how to validate a user-defined strategy (I guess we have some code somewhere that does it). Could you please let me know what function validates a user-defined strategy?

I guess first off we shouldn't call it validation :) we are just running unit tests here to show that the functionality is expected. By definition you can't test all user functions as the point of the mechanism is to allow them to do anything all we can do is test the mechanism itself. For the original impute function I apparently just did a single unit test where I created a new strategy called "XX" and assigned it equal to "JR and then tested that the results were the same which can be found [here](https://github.com/insightsengineering/rbmi/blob/develop/tests/testthat/test-fullusage.R#L336). I would recommend doing something similar just to show that the mechanism is working as expected

R/simulate_data.R
gowerc: @nociale , whats your reason for not wanting to set defaults in `simulate_data()` I'm thinking ideally we would not expose `simulate_test_data()` to users as to avoid any confusion.

nociale:
> @nociale
> 
> regarding
> 
> > The prob_ice1 argument is one of the arguments to specify the probability of the ICE for each visit. Currently it could be set as a number or a vector of length equal to the length of mu. If a vector, its last element specifies the probability that the ICE happens at the last visit, i.e. the first visit affected by the ICE will be the "visit after the last visit". In other words, the last element of prob_ice1 is useless and the result will be the same regardless of what value is chosen. So, I thought of removing it and let prob_ice1 be of length equal to the number of visits minus 1. However, it might be confusing for the user that he cannot specify a probability of the ICE for each visit. What do you think is best?
> 
> Yer I was wondering about this and found it confusing myself. I think it is much more intuitive for the probability to apply to the visit in question rather than the subsequent visit i.e. if I specified `c(0, 0.5, 0.7)` it would make more sense to me if Visit 0 had 0% ICE's, visit 1 had 50% ICE's and visit 2 had 70% ICES. I would then just put in the documentation that the first digit is ignored as baseline will always have 0% ICE's. I really don't like the last digit being ignored.

I agree that it is not nice that the last digit is ignored, and it would be a useless parameter set by the user. I think I still like the fact that if I specify `c(0.2, 0.5, 0.7)` then we simulate that the probability of ice1 immediately after baseline is 0.2 (which means that first visit affect by ice1 is visit1) etc... However I am happy to change as you said if you think that is more intuitive. @wolbersm Any thought on this?

wolbersm:
Hi @nociale, I have no strong preference either way but I think I find the current set-up more intuitive than the alternative. 
The way I think of it is that the value at the visit decides whether the ICE occurs ("right after" the current visit and hence affecting only value after the current visit). Since the value of the last element of the vector is ignored, we could allow both the current input or an input where the vector is shorter by 1.

gowerc:
@nociale , 

Where there any outstanding items to this or are we good to merge ?

R/simulate_data.R
nociale: @gowerc We need to decide whether to provide some defaults so that one can run `simulate_data()` without any argument. Currently one has to set at least number of patients, mean trajectories and covariance matrices, and `post_ice1_traj`. My idea is that if the user specifies parameters only in one arm, then if he just uses the default for the other arm he could obtain some non-sense data. I am happy to provide defaults to make it run easily, but we should be aware that non-sense things are a bit less controllable. One "in-between" solution could also be the following:
- the user needs to specify one argument, `pars_c` (so at least mean, covariance matrix and number of subjects in the control arm)
- by default, `pars_t = pars_c`
- by default, post_ice1_traj = "MAR"
This would simulate trajectories assuming no treatment effect.

I agree that can avoid exposing `simulate_test_data()`.

nociale:
@gowerc I have created a new exported function `get_example_data()` and I have unexported `test_simulate_data()` and `as_vcov()`. 

I am saying this in case you want to make a last review of this.

Best,
Alessandro

gowerc:
@nociale ,  Given we now have the `get_example_data` function do you thinking we should "un-export"  `simulate_test_data`  just thinking it might be confusing having 2 very similar functions being exported

nociale:
@gowerc I have unexported `simulate_test_data` in the advanced vignette, since the old version of the advanced_vignette made use of simulate_test_data and thus the tests were failing when unexporting this function in this branch.

PS: `as_vcov()` is still used in the new version of the advanced_vignette. Thus, I have not unexported it yet. However we could unexport it as well if we change the code of the advanced_vignette where this function is called (in the "delta" section).

gowerc:
@nociale ,

Sorry I dont understand `simulate_test_data` still appears to be exported:

image

nociale:
Sorry it was not clear. In the branch `advanced_vignette` I unexported this function https://github.com/insightsengineering/rbmi/pull/282/files.

gowerc:
Have merged the advanced vingete would you mind rerunning the document function to resovlve the merge conflict ?

PR #276, opened 2022-01-14
Parallel

### LINKED ISSUE #227: Parallelization of the bootstrap and jackknife
Parallelization of the bootstrap and jackknife in `draws()`.

gowerc:
closes #227


R/draws.R
```diff
@@ -195,37 +224,52 @@ draws.condmean <- function(data, data_ice = NULL, vars, method) {
...
-#' Bootstrapping refers to resampling of subjects (and their full longitudinal data) from `data` with
-#' replacement and and not to resampling of individual rows from `data`.
-#' If the model fit fails in the original dataset, an error message is thrown. If
-#' the model fit fails in a bootstrap sample, then that bootstrap sample is not considered
-#' and replaced by another bootstrap sample. `method$threshold` defines the
-#' maximum fraction of model fit failures: if the number of failures overcomes
-#' `ceiling(method$threshold * method$n_samples)` the process stops and an error message
-#' is displayed.
+#'
+#' This function takes a Stack object which contains multiple lists of patient ids. The function
```
nociale, line 242:
```suggestion
#' This function takes a `Stack` object which contains multiple lists of patient ids. The function
```

R/draws.R
```diff
@@ -195,37 +224,52 @@ draws.condmean <- function(data, data_ice = NULL, vars, method) {
...
+#' takes this Stack and pulls a set ids and then constructs a dataset just consisting of these
+#' patients (i.e. potentially a boostrap or a jackknife sample).
+#' 
+#' The function then fits a MMRM model to this dataset to create a sample object. The function
+#' repeats this process until `n_target_samples` have been reached. If more than `failure_liimt`
```
nociale, line 247:
```suggestion
#' repeats this process until `n_target_samples` have been reached. If more than `failure_limit`
```

R/stack.R
```diff
@@ -0,0 +1,76 @@
+#' R6 Class for a FIFO stack
+#'
+#' @description
+#'
+#' This is a simple stack object offering add / pop functionality
+#'
+#' @import R6
+#' @export
```
nociale, line 8: Just to understand, why did you choose a R6 class for this, and not a S3 method for jackknife and bootstrap?

gowerc: I used R6 for the stack object itself as it allows for in-place modification whilst still being able to return a value

```
x <- Stack()
val1 <- x$pop()
val2 <- x$pop()
```

vs something like

```
x <- Stack()

z1 <- pop(x)
x <- z1$x
val1 <- z1$vals

z2 <- pop(x)
x <- z2$x
val2 <- z2$vals
```

In terms of implementing the jackjnife and bootstrap I probabally could put an S3 wrapper around the Stack object which is pretty much what I've done [here](https://github.com/insightsengineering/rbmi/pull/276/files#diff-e26e6ab8b4b9c6e3658ed099e58fa294d8e8d7afa29f709ed5a4c66bcbefc5a0R49). In the end I opted not to because the bootstrap and condmean both have the same implementation so would need to have a 3rd function for them both to point to and it just got a bit messy.

PR #282, opened 2022-01-28
Advanced vignette

### LINKED ISSUE #229: Improvements of advanced vignette
- [x]  Use the improved data simulation function as per #225 
- [x] Better description and summary of when changing imputation strategy in `impute()` is allowed
- [x] Add section of example of usage with time-varying covariates. Show that MAR imputation here is equivalent to CIR imputation without time varying covariates (assuming post-ICE data are simulated according to CIR trajectory). Keep using Bayes as done in the rest of the vignette or condmean imputation?
- [x] Include realistic delta adjustement and tipping point analysis (look at Cro et al examples)

nociale:
Closes #229


nociale:
Hi @gowerc @wolbersm! This is a proposal of the improvements of the advanced vignette. I tried to make it as per the discussion we had on the issue #229. I am curious to hear what you think about it and modify accordingly (you can also modify it by yourself if you think it is more efficient).

Best regards,
Alessandro

nociale:
From @gowerc comments:

- [ ] `factor(ids, levels = unique(ids))` is done in 4 different locations. Perhaps its worth just applying this once at the start of the data generation.
- [ ] After ice1 the control group shouldn't have their outcome updated (even if from the same distribution).

gowerc:
@nociale , in the interest of time I will try and resolve a lot of these now but in the future please can you run the linter on the code :) 

image

gowerc:
@nociale ,  Additionally I replace one of your sections with inline r expressions to make it slightly easier to read:

image

vignettes/advanced.Rmd
```diff
@@ -26,102 +29,290 @@ knitr::opts_chunk$set(
 The purpose of this vignette is to provide an overview of some of the more advanced
 features of the `rbmi` package. 
 
+# Data simulation {#sec:dataSimul}
+
+In order to demonstrate the advanced functions we will first create a simulated dataset with the `rbmi` function `simulate_data()`.
+A reader interested in other sections of the vignette must not necessarly read through this section.
+The `simulate_data()` function generates data for a two-arms clinical trial with longitudinal continuous outcome and two intercurrent events (ICEs).
```
gowerc, line 36: Maybe misleading as its "up to two ICEs per subject" its not guarenteed to generate 2 ICEs  (and ambigious if you mean overall or per patient

vignettes/advanced.Rmd
```diff
@@ -26,102 +29,290 @@ knitr::opts_chunk$set(
...
+One intercurrent event (ICE1) may be thought of as a discontinuation from study treatment due to study drug or condition related (SDCR) reasons.
```
gowerc, line 37: "may be" ?  Surely that is what it is ?

gowerc: Perhaps, "The ICE1 intercurrent event represents discontinuation from ..."

vignettes/advanced.Rmd
```diff
@@ -26,102 +29,290 @@ knitr::opts_chunk$set(
...
+The other one (ICE2) may be thought of as discontinuation from study treatment due to uninformative study drop-out.
+For the purposes of this vignette we consider the following scenario which is similar to the simulation study used in @Wolbers2021.
+It is a trial of an active drug versus placebo with 100 subjects per group and 6 post-baseline assessments (bi-monthly visits until 12 months):
```
gowerc, line 40: Perhaps reword, it just doesn't seem qutie right, at least the opener of "it is a trial" doesn't read quite right to me.

vignettes/advanced.Rmd
```diff
@@ -26,102 +29,290 @@ knitr::opts_chunk$set(
...
+    c(
+        sd_intercept^2,
+        cor_slopeInter * sd_intercept * sd_slope,
+        cor_slopeInter * sd_intercept * sd_slope,
+        sd_slope^2
+    ),
+    ncol = 2
+)
+
+Sigma <- cbind(1, time / 12) %*% covRE %*% rbind(1, time / 12) + diag(sd_error^2, nrow = length(time))
```
gowerc, line 81: To me this just seems like quite a lot of code which I feel is likely to lose a reader unless they are actively running the code snippits in parallel.  Is it viabe to simplify this at all?  Or maybe instead just hand write the values instead of deriving them so its more explicit to the reader ?

vignettes/advanced.Rmd
```diff
@@ -26,102 +29,290 @@ knitr::opts_chunk$set(
...
+- `pars_c`: The simulation parameters of the control arm
+- `pars_t`: The simulation parameters of the intervention arm
+- `post_ice_traj`: Specify how observed outcomes occurring after ICE1 are simulated
+
+For more information about this function, please refer to the function documentation `help(simulate_data)`.
+
+# Management of post-ICE observed values handled with a reference-based assumption {#sec:postICEobs}
+
+The simulated trial explained in section \@ref(sec:dataSimul) assumes that outcomes in the active arm observed after the ICE "treatment discontinuation" follow the increments observed in the control arm.
+Thus the imputation of missing data in the active arm after treatment discontinuation might be performed under the CIR assumption. How `rbmi` deals with observed values after an ICE handled using a reference-based method, is to include them in the analysis model but exclude them from the imputation model.
```
gowerc, line 128: "might be" ?

vignettes/advanced.Rmd
```diff
@@ -26,102 +29,290 @@ knitr::opts_chunk$set(
...
+
+Let's first set-up the estimator: Suppose that imputation model has the mean change from baseline as the outcome, includes the treatment group, the (categorical) visit, treatment-by-visit interactions, the baseline outcome, and baseline outcome-by-visit interactions as covariates, and assumes a common unstructured covariance matrix in both groups.
```
gowerc, line 130: Perhaps convert into a bullet point list for clarity ?

vignettes/advanced.Rmd
```diff
@@ -26,102 +29,290 @@ knitr::opts_chunk$set(
...
+```{r}
+
+# Create data_ice: CIR strategy for each patient in the active arm who stops the treatment
+data_ice_CIR <- data %>%
+    group_by(id = factor(id, levels = unique(id))) %>%
+    mutate(
+      is_interv = group == "Intervention",
+      has_ice = any(ind_ice1 == 1)
+    ) %>%
+    filter(is_interv & has_ice) %>% # Select pts in active arm that disc from treatment
```
gowerc, line 143:
```suggestion
    filter(is_interv, has_ice) %>% # Select pts in active arm that disc from treatment
```

vignettes/advanced.Rmd
```diff
@@ -26,102 +29,290 @@ knitr::opts_chunk$set(
...
+
+Let's first set-up the estimator: Suppose that imputation model has the mean change from baseline as the outcome, includes the treatment group, the (categorical) visit, treatment-by-visit interactions, the baseline outcome, and baseline outcome-by-visit interactions as covariates, and assumes a common unstructured covariance matrix in both groups.
+The analysis model is ANCOVA that also adjusts for the baseline outcome value. Below we report the code to set the variables of the imputation and analysis models.
+For details on the basic set-up to call the key rbmi functions, please refer to the "quickstart" vignette (`vignette("quickstart")`).
+
+```{r}
+
+# Create data_ice: CIR strategy for each patient in the active arm who stops the treatment
+data_ice_CIR <- data %>%
+    group_by(id = factor(id, levels = unique(id))) %>%
```
gowerc, line 138: Generally bad practice to do a mutation inside of a group by statement, but anyway I think after the recent change to `simulate_data` this can be removed now ?

vignettes/advanced.Rmd
```diff
@@ -26,102 +29,290 @@ knitr::opts_chunk$set(
...
+    mutate(
```
gowerc, line 139: This mutate block feels redundant to me, why not just put the logic in the filter block ?

vignettes/advanced.Rmd
```diff
@@ -26,102 +29,290 @@ knitr::opts_chunk$set(
...
+        visit = visit[ind_ice1 == 1][1] # Select first visit affected by the ICE
+    )
+data_ice_CIR$strategy <- "CIR"
+
+# Compute change from baseline
+data$change <- data$outcome - data$outcome_bl
+
+# Remove rows corresponding to baseline visit
+data <- data[data$visit %in% levels(data$visit)[-1], ]
+data$visit <- factor(data$visit, levels <- unique(data$visit))
```
gowerc, line 154: Considering this is a vignette I think we should be consistent with our logic and try to stick to dplyr as much as possible (i.e. where possible avoid switching between base and dplyr to make it read cleaner)

vignettes/advanced.Rmd
```diff
@@ -26,102 +29,290 @@ knitr::opts_chunk$set(
...
+    covariates = c("visit*outcome_bl", "visit*group"),
+    strategy = "strategy"
+    )
+
+vars_an <- vars
+vars_an$covariates <- "outcome_bl"
+
+```
+
+Let's assume that the chosen multiple imputation (MI) method is the approximate Bayesian MI with 10 imputations that can be set with the function `method_approxbayes()` as follows:
```
gowerc, line 170:
```suggestion
Let's say that the chosen multiple imputation (MI) method is the approximate Bayesian MI with 10 imputations; this can be specified with the function `method_approxbayes()` as follows:
```

vignettes/advanced.Rmd
```diff
@@ -26,102 +29,290 @@ knitr::opts_chunk$set(
...
+
+```{r}
+method <- method_approxbayes(n_sample = 10)
+```
+
+We can now call sequentially the 4 key functions of `rbmi` to perform the multiple imputation.
```
gowerc, line 176:
```suggestion
We can now sequentially call the 4 key functions of `rbmi` to perform the multiple imputation.
```

vignettes/advanced.Rmd
```diff
@@ -26,102 +29,290 @@ knitr::opts_chunk$set(
...
+Please note that the management of post-discontinuation observed data is performed without additional complexity for the user.
+`draws()` automatically excludes post-ICE data handled with a reference-based method using information provided by the argument `data_ice`.
```
gowerc, line 178: > with a reference-based method

Was just thinking that we only keep "MAR" data. But we still exclude LMCF data but this isn't strictly a reference method is it ?

vignettes/advanced.Rmd
```diff
@@ -26,102 +29,290 @@ knitr::opts_chunk$set(
...
+`impute()` will impute only actual missing data in `data[[vars$outcome]]`.
```
gowerc, line 179: I'm guessing you were trying to highlight here that missing data in  the covariates is not handled ? If so is it worth just stating that a bit more explicitly otherwise this sentance seems almost a bit out of place to me.

vignettes/advanced.Rmd
```diff
@@ -26,102 +29,290 @@ knitr::opts_chunk$set(
...
+
+```{r}
+set.seed(123)
+draw_obj <- draws(
```
gowerc, line 183: You might want to use the `quiet=TRUE` argument here unless you wanted draw's progress logs appearing in the vignette ?

vignettes/advanced.Rmd
```diff
@@ -26,102 +29,290 @@ knitr::opts_chunk$set(
...
+    impute_obj_CIR,
+    vars = vars_an
+)
+
+pool(ana_obj_CIR)
 ```
 
+# Adding time-varying covariates in the imputation model
+
+`rbmi` allows to include time-varying covariates in the imputation model.
```
gowerc, line 205:
```suggestion
`rbmi` allows for the inclusion of time-varying covariates in the imputation model.
```

vignettes/advanced.Rmd
```diff
@@ -26,102 +29,290 @@ knitr::opts_chunk$set(
...
+For example, lets suppose that we want to implement an estimator equivalent to the one described in section \@ref(sec:postICEobs) with the following differences:
+
+- There is an additional time-varying covariate that represents the number of visits from the ICE treatment discontinuation. This is set equal to 0 up to the treatment discontinuation and to the time from treatment discontinuation
+at subsequent visits. Additionally a treatment by time-varying covariate interaction terms is included in the model. This variable allows to estimate the post-ICE trajectory in each arm directly from the data, without explicitly making apriori any (reference-based) assumption.
+- The imputation is performed under MAR.
```
gowerc, line 210: This just looks a bit weird given you have 1 super long sentance and 1 short one both in bullet points,  perhaps just have two short bullet points and extract the qualifying text to a paragraph beneath it ?

vignettes/advanced.Rmd
```diff
@@ -26,102 +29,290 @@ knitr::opts_chunk$set(
...
+
+```{r}
+vars_tv <- set_vars(
+    subjid = "id",
+    visit = "visit",
+    outcome = "change",
+    group = "group",
+    covariates = c("visit*outcome_bl", "visit*group", "time_from_ice1*group"),
+    strategy = "strategy"
+    )
```
gowerc, line 232:
```suggestion
)
```

vignettes/advanced.Rmd
```diff
@@ -26,102 +29,290 @@ knitr::opts_chunk$set(
...
+```
+
+We can now call sequentially the 4 key `rbmi` functions:
```
gowerc, line 235:
```suggestion
We now sequentially call the 4 key `rbmi` functions:
```

vignettes/advanced.Rmd
```diff
@@ -26,102 +29,290 @@ knitr::opts_chunk$set(
...
+draw_obj <- draws(
+    data = data,
+    data_ice = NULL, # if NULL, MAR is assumed for every missing data
+    vars = vars_tv,
+    method = method
+)
+
+impute_obj_tv <- impute(
+    draw_obj,
+    references = c("Control" = "Control", "Intervention" = "Intervention")
```
gowerc, line 247: Is this correct ? As in are both groups meant to reference themselves ?

vignettes/advanced.Rmd
```diff
@@ -26,102 +29,290 @@ knitr::opts_chunk$set(
...
+From a theoretical point of view, such updates are only sensible if the base imputation model and the data to which it is fitted are also applicable to the revised imputation strategy.
+For example, as described in `help("draws", "rbmi")`, observed post-ICE data are included in the imputation model for the MAR strategy but excluded for non-MAR strategies.
+For this reason, the imputation strategy cannot be changed from MAR to a non-MAR strategy in the presence of observed data after the ICE.
+Likewise, a change from a non-MAR to MAR in the presence of observed data after the ICE triggers a warning because the imputation model was not fitted to all relevant data and may be inefficient.
```
gowerc, line 272: I still wonder if we should be more explicit than just "inefficient" as people could thing this means inefficient in terms of computing resources rather than not using all of the available data.

vignettes/advanced.Rmd
```diff
@@ -268,15 +466,161 @@ pool(anl_obj)
 
 # Delta Adjustment
 
+A delta-adjustment refers to a shift in the distribution of the (unobserved) outcome values by a certain amount. This might be used to perform sensitivity analyses.
```
gowerc, line 469: Would it be better to say "imputed" rather than "unobserved"  as the unobserved values never actually change, we just don't know them ?

gowerc: Also "might be" ?

vignettes/advanced.Rmd
```diff
@@ -268,15 +466,161 @@ pool(anl_obj)
...
+In the context of multiple imputation (MI), this is typically done by adding a fixed amount to the imputed values used in the primary analysis.
+This implies that the delta-adjustment can be performed prior to the third stage of the MI process: The fit of the analysis model (`analyse()`).
+That is, there is no additional consistent computational time needed to perform this type of sensitivity analysis since there is no need to re-fit the imputation model.
```
gowerc, line 472: Not sure "consistent" is the right word here. Perhaps just re-phrase as, "this means this type of sensitivity analysis can be perforemed without having to re-fit the imputation model"

vignettes/advanced.Rmd
```diff
@@ -268,15 +466,161 @@ pool(anl_obj)
...
+For details, please see @CroEtAlTutorial2020.
+
+## Manually set `delta` {#sec:manualDelta}
+
+The `delta` argument of `analyse()` allows users to modify the outcome variable prior to the analysis.
+To do this, the user needs to provide a `data.frame` containing a column for the subject
```
gowerc, line 478: containing columns for **both** the subject and visit

vignettes/advanced.Rmd
```diff
@@ -268,15 +466,161 @@ pool(anl_obj)
...
+    group_by(visit) %>%
+    summarise(delta = mean(change, na.rm = TRUE))
+mean_chg
+```
+Note that `na.rm = TRUE` is used to ignore the missing data.
+
+Suppose that we want to apply a delta equal to the $20\%$ of the mean change from baseline, i.e.: 
+```{r} 
+deltas <- mean_chg
+deltas$delta <- 0.2 * mean_chg$delta
```
gowerc, line 507: Try to avoid mix and matching dplyr and base R. I think we should be consistant

vignettes/advanced.Rmd
```diff
@@ -268,15 +466,161 @@ pool(anl_obj)
...
+delta_adj <- function(impute_obj, mean_chg, prop) {
+
+    # Get delta template dataset
+    dat_delta <- delta_template(imputations = impute_obj)
+
+    # Estimate delta values
+    mean_chg$delta <- prop * mean_chg$delta
+
+    # Select observations corresponding to missing values after the ICE
+    index_delta <- dat_delta$is_missing & dat_delta$is_post_ice
```
gowerc, line 556: Again lets not mix base and dplyr

vignettes/advanced.Rmd
```diff
@@ -268,15 +466,161 @@ pool(anl_obj)
...
+```
+
+To perform the tipping point analysis, we loop over increasing values of the proportion and we stop when the p-value is above the desired confidence level (we assume that it is equal to 0.05).
+
+```{r}
+prop <- seq(from = 0.1, to = 1, by = 0.1)
+
+iter <- 1
+p_val <- 0
+while (p_val <= 0.05 & iter <= length(prop)) {
```
gowerc, line 583: I would avoid a while loop, I would simply just run all of the range and then call out the one that tripped the threshold. Only for the sole reason that while loops can be hard to get your head round at times for a reader.

gowerc (changes requested):
please address the above comments :)

vignettes/advanced.Rmd
wolbersm: I think it would be nice to hand-write some of this, e.g. time, muC, muT, and covRE. I would keep the formula for Sigma.

vignettes/advanced.Rmd
nociale, line 36:
```suggestion
The `simulate_data()` function generates data for a two-arms clinical trial with longitudinal continuous outcome and up to two intercurrent events (ICEs).
```

nociale: Agreed

vignettes/advanced.Rmd
nociale, line 40:
```suggestion
Let's assume a trial of an active drug versus placebo with 100 subjects per group and 6 post-baseline assessments (bi-monthly visits until 12 months):
```

nociale: "Let's consider a trial of ...."

vignettes/advanced.Rmd
nociale: ok, will do it

vignettes/advanced.Rmd
nociale: ""
```suggestion
Thus we impute missing data in the active arm after treatment discontinuation under the CIR assumption. How `rbmi` deals with observed values after an ICE handled using a reference-based method, is to include them in the analysis model but exclude them from the imputation model.
```

vignettes/advanced.Rmd
nociale: I think it is correct to exclude post-ICE data that follow a LMCF assumption, since this is not compatible with a MAR assumptions assumed by the base imputation model (MMRM)

vignettes/advanced.Rmd
nociale: I want to say that impute() will consider as non-missing those observed values that are artificially set to NA in draws() because they follow a non-MAR assumption... That is, impute() imputed only acutal missing data as provided in the dataset from the user. Happy to rewrite it

vignettes/advanced.Rmd
```diff
@@ -26,102 +29,290 @@ knitr::opts_chunk$set(
...
+pool(ana_obj_CIR)
+```
+
+# Adding time-varying covariates in the imputation model
+
+`rbmi` allows to include time-varying covariates in the imputation model.
+For example, lets suppose that we want to implement an estimator equivalent to the one described in section \@ref(sec:postICEobs) with the following differences:
+
+- There is an additional time-varying covariate that represents the number of visits from the ICE treatment discontinuation. This is set equal to 0 up to the treatment discontinuation and to the time from treatment discontinuation
+at subsequent visits. Additionally a treatment by time-varying covariate interaction terms is included in the model. This variable allows to estimate the post-ICE trajectory in each arm directly from the data, without explicitly making apriori any (reference-based) assumption.
```
nociale, line 209:
```suggestion
at subsequent visits. 
- Additionally a treatment by time-varying covariate interaction terms is included in the model. This variable allows to estimate the post-ICE trajectory in each arm directly from the data, without explicitly making apriori any (reference-based) assumption.
```

vignettes/advanced.Rmd
nociale: Ok, I separated the long sentence into 2 bullet points

vignettes/advanced.Rmd
nociale: Yes, I think this is correct, since we impute under MAR.

vignettes/advanced.Rmd
nociale, line 469:
```suggestion
A delta-adjustment refers to a shift in the distribution of the (imputed) outcome values by a certain amount. This can be used to perform sensitivity analyses.
```

vignettes/advanced.Rmd
nociale, line 472:
```suggestion
This means this type of sensitivity analysis can be perforemed without having to re-fit the imputation model
```

vignettes/advanced.Rmd
nociale, line 478:
```suggestion
To do this, the user needs to provide a `data.frame` containing columns for both the subject
```

vignettes/advanced.Rmd
nociale: mmm, strong preference on this? 

I think sometimes one could model similarly other type of intercurrent events through ICE1. ICE1 represents an ICE that happens more or less likely depending on the current outcome value. Sometimes this could also be e.g start of symptomatic treatment? Could it be used to model death in certain settings? (even though the post-ice1 trajectory is done under a reference-based method, this usually makes sense mainly for treatment discontinuation...)

vignettes/advanced.Rmd
```diff
@@ -26,102 +29,282 @@ knitr::opts_chunk$set(
...
+- `post_ice_traj`: Specify how observed outcomes occurring after ICE1 are simulated
+
+For more information about this function, please refer to the function documentation `help(simulate_data)`.
+
+# Management of post-ICE observed values handled with a reference-based assumption {#sec:postICEobs}
+
+The simulated trial explained in section \@ref(sec:dataSimul) assumes that outcomes in the active arm observed after the ICE "treatment discontinuation" follow the increments observed in the control arm.
+Thus the imputation of missing data in the active arm after treatment discontinuation might be performed under the CIR assumption. How `rbmi` deals with observed values after an ICE handled using a reference-based method, is to include them in the analysis model but exclude them from the imputation model.
+
+Let's first set-up the estimator: Suppose that imputation model has the mean change from baseline as the outcome, includes the treatment group, the (categorical) visit, treatment-by-visit interactions, the baseline outcome, and baseline outcome-by-visit interactions as covariates, and assumes a common unstructured covariance matrix in both groups.
```
nociale, line 122:
```suggestion
Let's first set-up the estimator: Suppose that imputation model has:
- The mean change from baseline as the outcome
- Includes the treatment group, the (categorical) visit, treatment-by-visit interactions, the baseline outcome, and baseline outcome-by-visit interactions as covariates
- Assumes a common unstructured covariance matrix in both groups.
```

vignettes/advanced.Rmd
```diff
@@ -26,102 +29,282 @@ knitr::opts_chunk$set(
...
+The analysis model is ANCOVA that also adjusts for the baseline outcome value. Below we report the code to set the variables of the imputation and analysis models.
+For details on the basic set-up to call the key rbmi functions, please refer to the "quickstart" vignette (`vignette("quickstart")`).
+
+```{r}
+
+# Create data_ice: CIR strategy for each patient in the active arm who stops the treatment
+data_ice_CIR <- data %>%
+    group_by(id = factor(id, levels = unique(id))) %>%
```
nociale, line 130:
```suggestion
    group_by(id) %>%
```

vignettes/advanced.Rmd
wolbersm: We could also just be neutral and say that the occurrence of ICE1 may depend on the observed outcome value and that the function allows to simulate post-ICE1 outcomes under reference-based assumptions. For example, ICE1 could be a a discontinuation from study treatment due to study drug or condition related (SDCR) reasons.

vignettes/advanced.Rmd
nociale: We might change this and allow user to not need to specify `references` when imputation is under MAR

vignettes/advanced.Rmd
nociale: We could just delete "..and may be inefficient"?

nociale: Even if maybe it is clear enough that here inefficiency is not related to computing resources..

vignettes/advanced.Rmd
nociale: Marcel wants to change this as well. I will not resolve this conversation and I will have a look again after his changes.

vignettes/advanced.Rmd
nociale: done

vignettes/advanced.Rmd
nociale: done

wolbersm:
Hi @nociale and @gowerc 

I have just completed a major review of the advanced vignette including quite a lot of editing and some substantially revised content (e.g. the tipping point analysis). It would be great if one or, ideally, both of you could have another read through the whole vignette to make sure that all is correct and well-written. Thanks!

vignettes/advanced.Rmd
```diff
@@ -20,152 +24,339 @@ knitr::opts_chunk$set(
...
+The sections of the vignette are relatively self-contained, i.e. readers should be able to jump directly to the section which covers the functionality that they are most interested in. 
 
+# Data simulation using function `simulate_data()` {#sec:dataSimul}
 
-In order to demonstrate these advanced functions we will first create a simulated dataset
-and perform a mock analysis so that all the require objects are available.
+In order to demonstrate the advanced functions we will first create a simulated dataset with the `rbmi` function `simulate_data()`.
+The `simulate_data()` function generates data from a randomized clinical trial with longitudinal continuous outcomes and up to two different types of intercurrent events (ICEs).
+One intercurrent event (ICE1) may be thought of as a discontinuation from study treatment due to study drug or condition related (SDCR) reasons.
+The other event (ICE2) may be thought of as discontinuation from study treatment due to not study drug or condition related (SDCR) reasons.
```
nociale, line 37:
```suggestion
The other event (ICE2) may be thought of as discontinuation from study treatment due to not study drug or condition related (NSDCR) reasons. 
```

vignettes/advanced.Rmd
```diff
@@ -20,152 +24,339 @@ knitr::opts_chunk$set(
...
 
+# Data simulation using function `simulate_data()` {#sec:dataSimul}
 
-In order to demonstrate these advanced functions we will first create a simulated dataset
-and perform a mock analysis so that all the require objects are available.
+In order to demonstrate the advanced functions we will first create a simulated dataset with the `rbmi` function `simulate_data()`.
+The `simulate_data()` function generates data from a randomized clinical trial with longitudinal continuous outcomes and up to two different types of intercurrent events (ICEs).
+One intercurrent event (ICE1) may be thought of as a discontinuation from study treatment due to study drug or condition related (SDCR) reasons.
+The other event (ICE2) may be thought of as discontinuation from study treatment due to not study drug or condition related (SDCR) reasons. 
+For the purpose of this vignette, we simulate data similarly to the simulation study reported in @Wolbers2021 (though we change some of the simulation parameters) and include only one  ICE type (ICE1).
```
nociale, line 38:
```suggestion
For the purpose of this vignette, we simulate data similarly to the simulation study reported in @Wolbers2021 (though we change some of the simulation parameters) and include only one ICE type (ICE1).
```

vignettes/advanced.Rmd
```diff
@@ -20,152 +24,339 @@ knitr::opts_chunk$set(
...
+    any_ICE=(any(ind_ice1==1)),
+    any_NA = any(is.na(outcome))) %>%
+  group_by(group) %>%
+  summarise(
+      subjects_with_ICE=sum(any_ICE),
+      subjects_with_missings=sum(any_NA)
+  )
 ```
 
+# Handling of observed post-ICE data in `rbmi` and imputation under CIR {#sec:postICEobs}
```
nociale, line 133: @wolbersm I think this section is general for reference-based imputation...
```suggestion
# Handling of observed post-ICE data in `rbmi` under reference-based imputation {#sec:postICEobs} 
```

vignettes/advanced.Rmd
```diff
@@ -20,152 +24,339 @@ knitr::opts_chunk$set(
...
+  summarise(
+    visit = visit[1], # Select first visit affected by the ICE
+    strategy = strategy[1])
+
+# Compute endpoint of interest: change from baseline
+data$change <- data$outcome - data$outcome_bl
+
+# Remove rows corresponding to baseline visits
+data <- data %>% filter(visit!="0")
+data$visit <- factor(data$visit, levels = unique(data$visit))
```
nociale, line 173:
```suggestion
# Compute endpoint of interest: change from baseline and
# remove rows corresponding to baseline visits
data <- data %>% 
    filter(visit != 0) %>% 
    mutate(
        change = outcome - outcome_bl,
        visit = factor(visit, levels = unique(visit))
    )
```

vignettes/advanced.Rmd
```diff
@@ -20,152 +24,341 @@ knitr::opts_chunk$set(
...
+The `draws()` function is by far the most computationally intensive function in `rbmi`.
+In some settings, it may be important to explore the impact of a change in the
+reference-based imputation strategy on the results.
+Such a change does not affect the imputation model but it does
+affect the subsequent imputation step. 
+In order to allow changes in the imputation strategy without having to re-run the 
+`draws()` function, the function `impute()` has an additional argument `update_strategies`.
+
+However, please note that this functionality comes with some important limitations:
+As described at the beginning of Section \@ref(sec:postICEobs), post-ICE outcomes are included in the input dataset  for the base imputation model if the imputation method is `MAR` but they are excluded for  reference-based imputation methods (such as `CIR`, `CR`, and `JR`).
```
nociale, line 240:
```suggestion
As described at the beginning of Section \@ref(sec:postICEobs), post-ICE outcomes are included in the input dataset  for the base imputation model if the imputation method is `MAR` but they are excluded for reference-based imputation methods (such as `CIR`, `CR`, and `JR`).
```

vignettes/advanced.Rmd
```diff
@@ -20,152 +24,341 @@ knitr::opts_chunk$set(
...
+Therefore, `updata_strategies` cannot be applied if the imputation strategy is changed from a MAR to a non-MAR strategy in the presence of observed post-ICE outcomes. Similarly, a change from a non-MAR strategy to MAR triggers a warning in the presence of observed post-ICE outcomes because the base imputation model was not fitted to all relevant data under MAR.
+Finally, `update_strategies` cannot be applied if the timing of any of the ICEs is changed (in argument `data_ice`) in addition to the imputation strategy.  
+
+As an example, we described an analysis under a copy increments in reference (CIR) assumption in the previous section. Let's assume we want to change this strategy to a jump to reference imputation strategy for a sensitivity analysis. This can be efficiently implemented using `update_strategies` as follows: 
+
+```{r}
+# Change ICE strategy from CIR to JR
+data_ice_JR <- data_ice_CIR
+data_ice_JR$strategy[data_ice_JR$strategy=="CIR"] <- "JR"
```
nociale, line 249:
```suggestion
data_ice_JR <- data_ice_CIR %>% 
    mutate(strategy = ifelse(strategy == "CIR", "JR", strategy))
```

vignettes/advanced.Rmd
```diff
@@ -20,152 +24,341 @@ knitr::opts_chunk$set(
...
 
 ```{r}
 strategy_CIR
 ```
 
-To further illustrate this for a simple example, assume that a strategy is to be implemented according to which the marginal mean of the imputation distribution is equal to the marginal mean trajectory for the subject according to their assigned group and covariates up to the ICE and equal to the average of the visit-wise marginal means based on the subjects covariates and the assigned group or the reference group, respectively. For the covariance matrix of the marginal imputation distribution, the covariance matrix from the assigned group is taken. 
+To further illustrate this for a simple example, assume that a strategy is to be implemented according to which the marginal mean of the imputation distribution is equal to the marginal mean trajectory for the subject according to their assigned group and covariates up to the ICE and equal to the average of the visit-wise marginal means based on the subjects covariates and the assigned group or the reference group, respectively.
+For the covariance matrix of the marginal imputation distribution, the covariance matrix from the assigned group is taken.
```
nociale, line 354: Could be better to split this long sentence into bullet points (e.g. as follows):
```suggestion
To further illustrate this for a simple example, assume that a new strategy is to be implemented as follows:
- The marginal mean of the imputation distribution is equal to the marginal mean trajectory for the subject according to their assigned group and covariates up to the ICE.
- After the ICE the marginal mean of the imputation distribution is equal to the average of the visit-wise marginal means based on the subjects covariates and the assigned group or the reference group, respectively.
- For the covariance matrix of the marginal imputation distribution, the covariance matrix from the assigned group is taken. 
```

vignettes/advanced.Rmd
```diff
@@ -189,108 +382,188 @@ index_mar <- c(TRUE, TRUE, FALSE)
...
-To incorporate this into `rbmi`, `data_ice` needs to be updated such that subjects are 
-specified as using the `AVG` imputation strategy. Additionally, the function needs
+To incorporate this into `rbmi`, `data_ice` needs to be updated such that the strategy `AVG` is specified for visits affected by the ICE. Additionally, the function needs
 to be provided to `impute()` via the `getStrategies()` function as shown below:
 
 ```{r}
-dat_ice$strategy <- "AVG"
-dat_ice
+data_ice_AVG <- data_ice_CIR
+data_ice_AVG$strategy[data_ice_AVG$strategy=="CIR"] <- "AVG"
```
nociale, line 390:
```suggestion
data_ice_AVG <- data_ice_CIR %>% 
    mutate(strategy = ifelse(strategy == "CIR", "AVG", strategy))
```

vignettes/advanced.Rmd
```diff
@@ -189,108 +382,188 @@ index_mar <- c(TRUE, TRUE, FALSE)
...
+pool_obj_prop <- pool(ana_obj_prop)
+pool_obj_prop
+
+# convert treatment effect estimate from log-OR to OR scale
+tmp <- as.data.frame(pool_obj_prop)
+tmp$OR <- exp(tmp$est); tmp$OR.lci <- exp(tmp$lci); tmp$OR.uci <- exp(tmp$uci)
+tmp[,c("parameter","OR","OR.lci","OR.uci")]
+```
+
+Note that if the user wants `rbmi` to use a normal approximation to the pooled test statistics, then the degrees of freedom need to be set to `df = NA` (as per the above example). If the degrees of freedom of the complete data test statistics are known or if the degrees of freedom are set to `df = Inf`, then `rbmi` pools the degrees of freedom accross imputed datasets according to the rule by Barnard and Rubin (see the "Statistical Specifications" vignette (`vignette("stat_specs", package = "rbmi")` for details). According to this rule, infinite degrees of freedom for the complete data analysis do not imply that the pooled degrees of freedom are also infinite.
```
nociale, line 479:
```suggestion
Note that if the user wants `rbmi` to use a normal approximation to the pooled test statistics, then the degrees of freedom need to be set to `df = NA` (as per the above example). If the degrees of freedom of the complete data test statistics are known or if the degrees of freedom are set to `df = Inf`, then `rbmi` pools the degrees of freedom accross imputed datasets according to the rule by Barnard and Rubin (see the "Statistical Specifications" vignette (`vignette("stat_specs", package = "rbmi")` for details). According to this rule, infinite degrees of freedom for the complete data analysis does not imply that the pooled degrees of freedom are also infinite.
```

vignettes/advanced.Rmd
```diff
@@ -189,108 +382,188 @@ index_mar <- c(TRUE, TRUE, FALSE)
...
+ana_obj_prop <- analyse(impute_obj_CIR, vars = vars_an, fun = compare_prop_lastvisit)
+
+# pool
+pool_obj_prop <- pool(ana_obj_prop)
+pool_obj_prop
+
+# convert treatment effect estimate from log-OR to OR scale
+tmp <- as.data.frame(pool_obj_prop)
+tmp$OR <- exp(tmp$est); tmp$OR.lci <- exp(tmp$lci); tmp$OR.uci <- exp(tmp$uci)
+tmp[,c("parameter","OR","OR.lci","OR.uci")]
```
nociale, line 476:
```suggestion
tmp[, c("parameter", "OR", "OR.lci", "OR.uci")]
```

vignettes/advanced.Rmd
```diff
@@ -20,152 +24,341 @@ knitr::opts_chunk$set(
...
+post_ice_traj <- "CIR"
+
+# Simulate data
+data <- simulate_data(
+    pars_c = parsC,
+    pars_t = parsT,
+    post_ice1_traj = post_ice_traj
+)
+
+str(data)
```
nociale, line 117: I personally like more "head(data)" but not strong preference
```suggestion
head(data)
```

vignettes/advanced.Rmd
```diff
@@ -20,152 +24,341 @@ knitr::opts_chunk$set(
...
-dat_ice_CIR$strategy <- "CIR"
-dat_ice_CIR
+# Create data_ice including the subject's first visit affected by the ICE and the imputation strategy
+# Imputation strategy for post-ICE data is CIR in the intervention group and MAR for the control group 
+# (note that ICEs which are handled using MAR are optional and do not impact the analysis
+#  because imputation of missing data under MAR is the default)
+data_ice_CIR <- data %>%
+  group_by(id) %>%
+  filter(ind_ice1==1) %>% # select visits with ICEs
+  mutate(strategy = ifelse(group == "Intervention","CIR","MAR")) %>%
```
nociale, line 163:
```suggestion
  mutate(strategy = ifelse(group == "Intervention", "CIR", "MAR")) %>%
```

gowerc:
Hey @wolbersm & @nociale , 

Just wanted to check where we are at with this PR, are either of you still actively working on it or am I ok to review / merge ?

wolbersm:
@gowerc I am no longer actively working on this.

nociale:
@gowerc I have also finished working on this branch. I think it would be nice if you could review the advanced vignette a last time (mostly related to the coding part) and then we can merge

PR #283, opened 2022-02-02
Update lsmeans to use "proportional"

### LINKED ISSUE #275: Review LSmeans implementaiton
Potentially a need for lsmeans to be weighted by data prevalenance

i.e. 
    1. calculate prediction for entire dataset setting treatment to "placebo"
    2. calculate mean of this prediction
    3. calculate prediction for entire dataset setting treatment to "active"
    4. calculate mean of this prediciton

gowerc:
Closes #275

Hi @wolbersm  & @nociale ,

This PR has implemented "proportional" lsmeans and sets it to be the default. Users can use "equal" weighting (which is the SAS and emmeans default) by specifying `anocva(..., weights = "equal")`.

Only thing I'm not confident on is if I've provided enough detail in the user facing documentation (see `ancova()`). Please take a look and let me know what you think.


R/ancova.R
```diff
@@ -12,6 +12,8 @@
 #' fit the ancova model at. If `NULL`, a separate ancova model will be fit to the
 #' outcomes from each visits (as determined by `unique(data[[vars$visit]])`).
 #' See details.
+#' @param weights Character, specifies wether to use "proportional" or "equal" weighting for each
+#' categorical covariate combination when calculating the lsmeans.
```
wolbersm, line 16: Suggest to add the following to the help: 
"proportional" is the default and is equivalent to standardization, i.e. lsmeans in each treatment group are equal to the predicted mean outcome from the ancova model for that treatment group based on baseline characteristics of all subjects regardless of their assigned treatment.

R/ancova.R
```diff
@@ -12,6 +12,8 @@
 #' fit the ancova model at. If `NULL`, a separate ancova model will be fit to the
 #' outcomes from each visits (as determined by `unique(data[[vars$visit]])`).
 #' See details.
+#' @param weights Character, specifies wether to use "proportional" or "equal" weighting for each
```
nociale, line 15: I think we should describe a bit what "'proportional" and "equal" weighting are. For "proportional" I think Marcel commented below. For "equal" weighting I would also add a short sentence.

nociale:
@gowerc I have added a test made by @wolbersm where we check that lsmean (proportional) for a treatment group is equal to predicting the outcome for all patients assuming they belong to a treatment group and then averaging. rbmi does the opposite: we "average" the design matrix and then we predict. The test verifies that for ancova (linear models) these two approaches are the same.

Craig, could you also consider whether it is more tidy to have a separate file "test-lsmeans" for lsmeans related testing? Currently everything is in test-ancova. I am fine with both, just feel free to choose what do you think it is best.

nociale (approved):

R/ancova.R
gowerc: @nociale  & @wolbersm ,  how about the following:

```
@param weights Character, either "proportional" (default) or "equal". Specifies the
weighting strategy to be used for categorical covariates when calculating the lsmeans.
See details.
```

and then:

```
 Weighting:

       "proportional" is the default scheme that is used. This is
       equivalent to standardization, i.e. the lsmeans in each group
       are equal to the predicted mean outcome from the ancova model
       for that group based on baseline characteristics of all subjects
       regardless of their assigned group. The alternative weighting
       scheme, "equal", creates hypothetical patients by expanding out
       all combinations of the models categorical covariates. The
       lsmean is then calculated as the average of the predicted mean
       outcome for these patients assuming they come from each of the
       groups in turn.

       In short:

         • "equal" weights categorical covariates equally across all
           theoretical combinations

         • "proportional" weights categorical covariates based upon
           their frequency of occourance in the data

```

(added in my most recent commit)

wolbersm: Thanks @gowerc. Looks good to me except for "occourance" -> "occurrence".

gowerc:
@nociale , I think I've addressed everything now and am going to merge this. Please create an issue if there is something you think i've missed / need to address still

PR #285, opened 2022-02-04
parallelize jacknife

harriscw:
Hi,

Just wondering if you thought something like this would make sense as jackknife is a parallelizeable procedure.  I played around a bit and assuming I did it right got some substantial speed up.  If you think this makes sense I'm thinking soemthing similar could be done with the bootstrap method.

Thanks

### DIFF
--- a/R/draws.R
+++ b/R/draws.R
@@ -172,13 +172,13 @@ draws.approxbayes <- function(data, data_ice = NULL, vars, method) {
 
 #' @rdname draws
 #' @export
-draws.condmean <- function(data, data_ice = NULL, vars, method) {
+draws.condmean <- function(data, data_ice = NULL, vars, method, mc.cores=1) {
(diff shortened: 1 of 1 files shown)


gowerc:
Hi @harriscw ,

Thanks for the PR.  I think I will have to reject this though as we have already implemented paralisation of the jack knife (as well as the conditional mean + approx bayesian) in our develop branch which we will hopefully be merging into the master branch in the next few weeks.  The PR in question can be found here: https://github.com/insightsengineering/rbmi/pull/276

An example of usage is:

```
draws(
    data = dat,
    data_ice = dat_ice,
    vars = vars,
    method = method_condmean(type= "jackknife"),
    ncores = 8
)
```

harriscw:
OK that sounds great, thanks @gowerc .  I'll install the package from the `develop` branch and play around.  Thanks!

PR #289, opened 2022-02-16
Update publisher permissions

dinakar29:
So that the bot is able to publish to GH pages.


dinakar29:
Associated docs: https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs#overview

PR #291, opened 2022-02-18
Allow default references when imputation is under MAR

### LINKED ISSUE #284: Allow missing references argument in impute() in case there are no ICEs or all ICEs have strategy MAR.
I think it's important that rhmi is also easy to use for pure MAR-imputation and in this case, I feel it is potentially confusing to require a "references" argument in impute(). 
I hope it's easy to fix this? If yes, it would be great to have this in the CRAN release.

nociale:
Closes #284

Now the default `references` argument is `NULL`. If `references` is `NULL` we internally check that the imputation strategy is `MAR` for all the subjects. If yes, `references` is automatically set to `c("A" = "A", "B" = "B",...)`. Otherwise we throw an error.


tests/testthat/test-impute.R
```diff
@@ -141,6 +141,75 @@ test_that("Basic Usage", {
...
+    drawsObj <- draws(dat, data_ice, vars, method, quiet = TRUE)
+    imputeObj <- impute(drawsObj)
+
+    expected_ref <- c(
+        "Control" = "Control",
+        "Intervention" = "Intervention"
+    )
+    expected_ref <- add_class(expected_ref, "references")
+
+    expect_equal(
```
gowerc, line 187: would it be more meaningful to re-impute and show that the imputations return identical objects?

gowerc:
2 minor comments but otherwise LGTM. Will pass review so you can merge once you are happy.

gowerc (approved):

gowerc:
@nociale , would you mind merging in the `main` branch. I just added a new cicd check.

PR #308, opened 2022-02-28
added warning to fill_locf

### LINKED ISSUE #305: expand_locf does not fill missing values in case first assessment is missing
- [ ] Add in documentation about this issue
- [ ] Throw a warning in case a first assessment is missing for some subject

gowerc:
Closes #305

Example of warning:

```
> df_actual <- fill_locf(
+         input_df,
+         vars = c("v1", "myfac1"),
+         group = "pt",
+         order = c("srt1", "srt2")
+     )
Warning message:
In fill_locf(input_df, vars = c("v1", "myfac1"), group = "pt", order = c("srt1",  :
  The following variables have missing values as their first value in one of more groups: `v1`, `myfac1`
 Please consult the man page for `fill_locf()` for further details / recommendations
```

@wolbersm , do you think this is sufficient with what you had in mind ?

### DIFF
--- a/R/expand.R
+++ b/R/expand.R
@@ -59,7 +59,7 @@ locf <- function(x) {
 #' ## Missing First Values
 #'
 #' The `fill_locf()` function performs last observation carried forward imputation.
-#' A natural consquence of this is that it is unable to impute missing observations if the
+#' A natural consequence of this is that it is unable to impute missing observations if the
 #' observation is the first value.
 #' A common request is for this function to impute the first value if it is missing
 #' however doing so risks silent errors in the case of time varying covariates.
--- a/R/expand.R
+++ b/R/expand.R
@@ -60,11 +60,12 @@ locf <- function(x) {
 #'
 #' The `fill_locf()` function performs last observation carried forward imputation.
 #' A natural consequence of this is that it is unable to impute missing observations if the
-#' observation is the first value.
-#' A common request is for this function to impute the first value if it is missing
-#' however doing so risks silent errors in the case of time varying covariates.
-#' The current recommendation for dealing with this problem is to first use `expand_locf()` on just
-#' the visits and time varying covariates and then merge on the baseline covariates afterwards i.e.
+#' observation is the first value for a given subject / grouping.
+#' These values are deliberately not imputed as doing so risks silent errors in the case of time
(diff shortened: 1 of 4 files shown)


wolbersm:
Hi @gowerc 
Sounds ok but I think it would be important to also say here and/or in the man page that a likely reason for this warning is that a row corresponding to the first visit is missing for at least one subjects in the input dataset. [And say what the workarounds are.]

R/expand.R
```diff
@@ -56,6 +56,29 @@ locf <- function(x) {
 #' `expand_locf()` a simple composition function of `fill_locf()` and `expand()` i.e.
 #' `fill_locf(expand(...))`.
 #'
+#' ## Missing First Values
+#'
+#' The `fill_locf()` function performs last observation carried forward imputation.
+#' A natural consquence of this is that it is unable to impute missing observations if the
```
nociale, line 62:
```suggestion
#' A natural consequence of this is that it is unable to impute missing observations if the
```

R/expand.R
```diff
@@ -56,6 +56,29 @@ locf <- function(x) {
...
+#' observation is the first value.
+#' A common request is for this function to impute the first value if it is missing
+#' however doing so risks silent errors in the case of time varying covariates.
```
nociale, line 65:
```suggestion
#' A common request for this function is to impute the value of a baseline covariate if the row corresponding to the first observation for a subject is missing.
#' However doing so risks silent errors in the case of time varying covariates, so it is not allowed.
#' Perhaps the easiest in practice is to use a LOCF imputation with the baseline visit included (which should never be a missing row) and then remove the baseline afterwards (in case the change from baseline is modeled).
```

nociale (approved):

nociale:
@gowerc A couple of comments for the documentation but otherwise all good.

R/expand.R
gowerc: @nociale , Sorry I'm not convinced by this wording. Was there a particular message you were trying to convey ?  In general in documentation I feel like we should try to avoid words like "perhaps", "maybe" , "potentially" and instead make it seem more authoritative.

nociale: I agree "perhaps" should be avoided

Issue #341, opened 2022-04-14
Documentation enhancements


nociale:
- In description of “draws” function: “The imputation model is a mixed effects model repeated measures (MMRM) model.” Remove first instance of the word model.
- In description of “draws” function: “It can be fit using frequentist maximum likelihood (ML) or restricted ML (REML)”. REML is also frequentist.
- In quickstart vignette: “The analysis model is an ANCOVA model with the treatment group as the primary covariate and adjustment for the baseline HAMD17 score.” This suggests group is the covariate while it is the factor.

PR #342, opened 2022-04-14
small changes in draws doc and quickstart vignette

nociale:
Addresses #341.


R/draws.R
```diff
@@ -9,9 +9,9 @@
 #' in the absence of intercurrent events (ICEs) handled using reference-based imputation methods.
 #' For this reason, any observed outcome data after ICEs, for which reference-based imputation methods are
 #' specified, are removed and considered as missing for the purpose of estimating the imputation model, and for
-#' this purpose only. The imputation model is a mixed effects model repeated measures (MMRM) model that is valid
+#' this purpose only. The imputation model is a mixed model for repeated measures (MMRM) that is valid
```
gowerc, line 12: Is "for" needed here? I thought it was just "Mixed model repeated measures" ?

vignettes/quickstart.Rmd
```diff
@@ -49,7 +49,7 @@ data("antidepressant_data")
...
+We consider an imputation model with the mean change from baseline in the HAMD17 score as the outcome (variable `CHANGE` in the dataset). The following covariates are included in the imputation model: the treatment group (`THERAPY`), the (categorical) visit (`VISIT`), treatment-by-visit interactions, the baseline HAMD17 score (`BASVAL`), and baseline HAMD17 score-by-visit interactions. A common unstructured covariance matrix structure is assumed for both groups. The analysis model is an ANCOVA model with the treatment group as the primary factor and adjustment for the baseline HAMD17 score.
```
gowerc, line 52: Not sure if I like this,  as in yes technically from an implementation point of view it is stored as a factor but thats not what we are trying to draw attention to here is it? As in isn't the point of this sentence trying to say that treatment group is the main variable of interest, I.e. the variable we are drawing inference on?

gowerc:
@nociale ,

Just to say that the vignettes no longer automatically update and need to be manually re-run. I added a section to the contributing guide on how to do this. Please let me know if its not clear so I can adjust it :)

R/draws.R
nociale: I think that this is also what more commonly you find in the literature...

vignettes/quickstart.Rmd
nociale: I think that calling the treatment group the "primary factor" in an ANCOVA model is in line with the wording of the theory of standard ANOVA (e.g. https://en.wikipedia.org/wiki/Analysis_of_variance#For_a_single_factor). The previous wording wasn't unclear to the JOSS reviewers.

gowerc:
LGTM

gowerc (approved):

PR #343, opened 2022-04-27
255

pengguanya:
This is a first attempt to modify the data-structure and add explicit visit information (as a column in a data.frame) for the result in pool object. A generic wrapper is used to modify the pool function instead of directly editing its code, as this type of changes might have a general interest, not specifically for only one function. 

Without knowing the detailed specification, this change does not intend to meet all expectations. Some questions are added under the discussion of the issue to get further specifications.

### DIFF
--- a/R/analyse.R
+++ b/R/analyse.R
@@ -341,6 +341,16 @@ print.analysis <- function(x, ...) {
         n_samp
     )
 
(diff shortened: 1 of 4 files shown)


pengguanya:
This update is outdated. See new pull request for branch 255-01

PR #352, opened 2022-05-06
Update analysis function

pengguanya:
@gowerc 

Functions modified
* analysis
* print.analysis
* aconva
* aconva_single

New functions added
* analysis.R/analysis_result (exported)
* analysis.R/as_analysis_result (exported)
* analysis.R/is.analysis_result (exported)
* analysis.R/namechecker
* analysis.R/ana_name_chker
* analysis.R/analysis_info
* utilities.R/add_meta

A new 'analysis_result' class has been defined. Both built-in analysis function ancova and customer analysis function take input such as
```
 x <- list(
        analysis_result(
            name = 'trt1',
            est = coef(mod)[[group]],
            se = sqrt(vcov(mod)[group, group]),
            df = df.residual(mod),
            meta = add_meta('visit', ...)
        ),
        analysis_result(
            name = 'trt2',
            est = coef(mod)[[group]],
            se = sqrt(vcov(mod)[group, group]),
            df = df.residual(mod),
            meta = list(visit='1', something_else='c')
        ),
        analysis_result(
            name = 'trt3',
            est = coef(mod)[[group]],
            se = sqrt(vcov(mod)[group, group]),
            df = df.residual(mod)
            # meta is optional
        )
    )
``` 

The benefits of defining a new class to me are
* Make it more explicit when user defining a analysis function
* Make it possible to define methods and generic functions associated to this class (more flexible to add attraction and improve internal interfaces)
* The base type of the class is implemented using list, therefore the lower level structure of the object does not change which leads to minimal impact to the internal interfaces

Various functions are implemented to validate the object, add meta information and extract information for analysis result. The analysis information are printed now in a tabular format with meta information in separate columns.

Tests have not been done yet. want to know your input and advice first. Thanks

analysis_info


gowerc:
Hi @pengguanya ,

On the surface level this looks good, as in the input / output is roughly what I was hoping for.  Only thing I can't make up my mind on is wether `meta` should be its own object or to just accept a base list. I think I am leaning towards having a list... 

With regards to the internal implementation though I think it might need to be altered slightly, for a start we made an active decision to not use the tidyverse within this package to try and keep the dependency chain as light as possible so you will need to remove all uses of dplyr/purrr/rlang.

Minor point but in terms of consistency with the result of the package please can you replace `stop_if_not` with `assert_that`

pengguanya:
@gowerc Thanks. I will make the changes and add test to each function for next pull request

As for now `meta` is a base list. The util function `add_meta` is only used to ease creation of `meta` for internal interfaces. It does return a base list and is not exported. So, for user, they need to do something like `meta = list(...)`

pengguanya:
This pull request is deprecated. Please review pull request #362 255 02 instead

Issue #274, opened 2022-01-10
Returning frequentist MMRM model object


gowerc:
Hi @wolbersm ,

We have been asked a few times by Paul to add return the model object of the MMRM fit for the bootstrap methods  for diagnostic / residual purposes.   

Talking to @nociale our current proposal would be that we return the MMRM fit to the original dataset (only) for the condmean jackknife + bootstrap methods,   for all other methods  (bayesian, approx bayes, blmi) we would carry on returning nothing.

What do you think ?

wolbersm:
We need to fit an MMRM model to the original data for all methods, right? 
Therefore, I'd suggest to return the original MMRM fit for all methods. 
Does this work or would it cause extra complications.

One totally separate issue is that the MMRM to the original data does exclude "non-MAR" post-ICE data but I guess this is unavoidable. Also, it's not 100% clear to me how to do good residual diagnostics for an MMRM model but I leave that to Paul for now.

wolbersm:
Thanks for clarifying @gowerc 
I am not 100% sure what Paul wants to do with the MMRM fit in any case but after some consideration, your original proposal sounds reasonable.

nociale:
@gowerc @wolbersm Just to be clear, currently `drawsObj$fit` is `NULL` for all the methods except `bayes` for which the `stanfit` object is returned. We could return the fit object from the `glmmTMB()` function for the MMRM fit on the original dataset for the `condmean` method (either bootstrap or jaccknife). `drawsObj$fit` would still be `NULL` for `bmlmi` and `approxbayes`. Even though we do fit MMRM on the original dataset also when these methods are called, in these cases (as for `bayes`) this fit serves only for initialization / setting starting values, so I would personally not output this fit object.

wolbersm:
Thanks, as long as the content of `drawsObj$fit` is properly documented, this sounds ok for me.

gowerc:
I'm not sure we would be able to document it other than saying its a "glmfit" object. Otherwise we have a moving target that our documentation is dependent on another teams implementation details.

wolbersm:
Yes, I think it's fine to just say that this contains the result of the imputation model fit returned by glmmTMB() or stan().

delmarp:
sounds good. I would also return the full object from the ancova function for the same reason to check residuals

pengguanya:
Just add here the relevant functions in case I forgot. There is a kind of chain relation among them
```
1. draws.condmean
2. get_draws_mle
[...]

PR #369, opened 2022-09-09
274 Return frequentist MMRM model object

pengguanya:
https://github.com/insightsengineering/rbmi/issues/274

Fetch model for condmean. All tests past at local

### DIFF
--- a/R/draws.R
+++ b/R/draws.R
@@ -190,7 +190,7 @@ draws.approxbayes <- function(data, data_ice = NULL, vars, method, ncores = 1, q
         ncores = ncores,
         n_target_samples = method$n_samples,
         failure_limit = ceiling(method$threshold * method$n_samples),
-        quiet = quiet,
+        quiet = quiet
     )
     return(x)
 }
--- a/R/draws.R
+++ b/R/draws.R
@@ -391,16 +391,10 @@ get_draws_mle <- function(
         }
     }
 
-    ret <- as_draws(
-        method = method,
-        samples = sample_list(samples),
-        data = longdata,
-        formula = longdata$formula,
-        n_failures = n_failed_samples
-    )
-    return(ret)
-}
+    args$samples <- sample_list(samples)
 
+    return(do.call(as_draws, args))
+}
 
 
 #' Fit MMRM and returns parameter estimates
--- a/R/draws.R
+++ b/R/draws.R
@@ -381,7 +381,17 @@ get_draws_mle <- function(
         msg = "Incorrect number of samples were produced"
     )
 
+    args <- list(
+        method = method,
+        samples = NULL,
+        data = longdata,
+        formula = longdata$formula,
+        n_failures = n_failed_samples
+    )
+
     if (first_sample_orig) {
+        args[['fit']] <- attr(initial_sample, 'model')
+        attr(initial_sample, 'model') <- NULL
         samples <- append(list(initial_sample), samples)
     }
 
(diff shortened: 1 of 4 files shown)


gowerc:
Have incorporated the update to use mmrm instead of glmmTMB. 

I also decided to convert this from using attributes to instead using environments to store the model object. Whilst it is a bit of an anti pattern I think it simplifies the code quite a bit and avoids the need redefining attributes across multiple locations.

I still need to update the documentation and unit tests but @nociale  would you mind skimming the changes to see what you think ?

nociale:
@gowerc It looks good to me, as if I do `drawObj$fit` I would get the fit object, and this is exactly what this branch should add. I don't see any problem in your solution! What about tests? Should we add a test that verifies that a `mmrm` object is actually returned?

nociale (approved):

gowerc:
@nociale 

Urgh, a big issue with this that I only just realised is that we completely do away with the variable names and also manually expand out dummy variables i.e:

image

Can't imagine people will be happy getting  `V1` / `V2` / `V3`...

The amount of code required to map that all bag to meaningful variable names I think might be a huge pain...  Not really sure what to do here :(

gowerc:
Some notes as to why this is difficult:

- We manually expand out the design matrix before handing it over to mmrm. We do this for 2 reasons (1) so that we can ensure missing values in the outcome aren't dropped (as we will impute them later) and (2) so that we can retain the design matrix so we can multiply it by the corresponding coefficients later in our processing. 

- We reset the names of all covariates to `Vx` as to avoid name collisions when we insert the grouping variables for the covariance matrix. In essence our formula becomes   `outcome ~ V1 + V2 + V3 + V4 + us(visit | id / group)`  where the first few variables (i.e. `V1-V3` ) are usually the expanded visit columns. 

We probably could re-write the model to be based on the original variable names and then just assume the expanded design matricies are properly aligned, but this would mean re-writing a non-trivial amount of code + unit tests.

gowerc:
Some more notes encase we did want to make this change.

the dataframe is expanded into a design matrix in `get_mmrm_sample()`.  It is then handed over to `fit_mmrm()` where the `mmrm::mmrm()` is called.   The columns are actually renamed twice, once in `as_model_df()` and again in `as_mmrm_df()`.

If we were to pass the raw dataframe to `mmrm::mmrm()` I think we could just delete `as_mmrm_df()`. Main thing would be to ensure the order of the coeficients returned by `mmrm::mmrm()` is consist with how the designmatrix is created in `impute()` in particular these two lines:

```
    dat_pt_mod <- as_model_df(dat_pt, data$formula)
    dat_ref_mod <- as_model_df(dat_ref, data$formula)
```

I don't think `as_model_df()` is used anywhere else. Looking at this the biggest issue I think by far would be updating the unit tests to account for all the refactoring of the model functions.

I can't decide if this is worth the hassle or not at the moment :s

nociale:
@gowerc That was a really good point. If I blind myself from considering the amount of work needed to complete this task, I would say that if we return the MMRM object from the base imputation model, we should keep the same variable names as in `data`. From a user perspective, looking at the coefficient of "V1", "V2" etc.. is quite meaningless.  So, if it is possible to ensure that the order of the variables is the same as in `impute`, then I would go for it. However since this new feature is not of primary importance for the package, I would also evaluate how much this is worth (maybe also based on feedbacks from user experiences etc..).

In short, I would personally add this feature only once the name of the variables is the same as in the input dataset.

nociale:
Plus, please see #382 which also applies here in principle..

PR #372, opened 2022-10-13
This PR aims to switch out glmmTMB and replace it with the MMRM package

gowerc:
This is currently a WIP.

- Main functionality swap has been done and all unit tests are passing (including full test suite)
- Am currently just cleaning up code and updating documentation regarding the glmmTMB package


R/mmrm.R
```diff
@@ -248,117 +222,52 @@ fit_mmrm <- function(
             simplify = FALSE
         )
     }
-    names(params$sigma) <- levels(group)
+    names(params$sigma) <- levels(group) # TODO test this assumption is violated
```
nociale, line 225: It seems correct.

I tried to run `mmrm` with separate covariance and `names(params$sigma)` was equal to `levels(group)`. I then tried to to re-factor `group` by changing the order of factors and then run again `mmrm`. The order of `names(params$sigma)`, as well as of the covariance matrices, was correctly inverted...

gowerc: thanks for checking, the commit I am currently working on also includes a unit test for this as well :)

CONTRIBUTING.md
```diff
@@ -87,19 +87,12 @@ A particular issue with testing this package is reproducibility. For the most pa
...
+new but has so far proven to be very stable, fast and reliable. If you do spot any issues
+with the MMRM package please do raise them in the corresponding GitHub Repository - [link](https://github.com/openpharma/mmrm/issues)
 
-The most common issue is a warning that `glmmTMB` has been compiled against a different version of `TMB` or `Matrix` compared to what is  installed locally. The only main to this is to manually recompile the package yourself by running:
-
-```
-install.packages(c("TMB", "glmmTMB"), type = "source")
-``` 
-
-Note that you will need to have [rtools](https://cran.r-project.org/bin/windows/Rtools/rtools42/rtools.html) installed if you are on a Windows machine or [Xcode](https://apps.apple.com/us/app/xcode/id497799835?mt=12) if you are running macOS (or somehow else have access to a C/C++ compiler).
```
danielinteractive, line 100: I think this would be good to keep, also applies here

NEWS.md
```diff
@@ -1,6 +1,8 @@
 # rbmi (development version)
 
-* No changes
+# rbmi 1.2.0
+
+* Replaced our `glmmTMB` dependency with the `mmrm` package. This has resulted in the package being more stable (less model fitting convergance issues) as well as reducing run times to by about 2/3rds.
```
danielinteractive, line 5:
```suggestion
* Replaced our `glmmTMB` dependency with the `mmrm` package. This has resulted in the package being more stable (less model fitting convergence issues) as well as speeding up run times 3-fold. 
```

R/mmrm.R
```diff
@@ -123,33 +115,30 @@ as_mmrm_formula <- function(mmrm_df, cov_struct) {
...
+    beta <- coef(fit)
     names(beta) <- NULL
 
-    sigma <- VarCorr(fit)$cond
-    sigma <- lapply(sigma, function(x) {
-        x <- as.matrix(as.data.frame(x))
-        colnames(x) <- NULL
-        rownames(x) <- NULL
-        return(x)
+    sigma <- fit$cov
```
danielinteractive, line 129:
```suggestion
    sigma <- VarCorr(fit)
```
should work as well here

danielinteractive (commented):
thanks a lot @gowerc , see few comments inside

CONTRIBUTING.md
gowerc: Ah, yes you still use `TMB` so yer this issue has not fully gone away. Good spot will add back in.

.github/workflows/build_docker.yaml
```diff
@@ -21,13 +21,13 @@ jobs:
               src_image: "rocker/r-ver:4.0.4", 
               cran_url: "https://packagemanager.rstudio.com/cran/2021-04-20",
               tag: "rbmi:r404",
-              mmrm_version : "v0.13"
+              mmrm_version : "v0.1.3"
```
danielinteractive, line 24: can already increase to v0.1.5 now - just had some minor fixes

gowerc: ty will update (this will be removed once mmrm hits the public rspm servers)

PR #419, opened 2024-08-02
Documentation to inform users about Rhat warning

### LINKED ISSUE #288: Warning message "The largest R-hat is 2.12, indicating chains have not mixed" often appears for low `n_samples`
It might be worth inquiring about this at a suitable discussion forum, e.g. at https://discourse.mc-stan.org/. First we need to check whether similar issue is already posted somewhere, and if not prepare a minimal example where this happens.

nociale:
Closes #288

### DIFF
--- a/R/draws.R
+++ b/R/draws.R
@@ -133,7 +133,7 @@
 #' - `group`: name of the column in `data` which contains the group variable.
 #' - `outcome`: name of the column in `data` which contains the outcome variable.
 #' - `covariates`: vector of characters which contains the covariates to be included
-#'   in the model (including interactions which are specified as `"covariateName1*covariateName2"``).
+#'   in the model (including interactions which are specified as `"covariateName1*covariateName2"`).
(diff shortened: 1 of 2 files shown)


nociale:
Hi @gowerc, I think your proposal could be implemented by setting the `thin` argument of `stan()` call to the default `1`, doing `burn_in + burn_between*n_samples` with the thinning of 1, and then select only the samples every `burn_between` (after the burn_in). Was this your idea?

This would maybe get rid of the warning because the diagnostics are run in a larger amount of MCMC iterations, but it would be overriding the choice of stan to run the diagnostics on the `n_samples` only. 

Despite I do think this would be an elegant solution to the problem, I would maybe still go with the note to the warning message in the documentation looks as it is an easy and reasonable solution. With a sufficiently high number of samples, the warning should disappear anyway.

I am tagging @wolbersm as well as we discussed this together earlier today.

wolbersm:
Thanks @nociale. As discussed, this reflects my opinion on the topic as well.

PR #431, opened 2024-09-27
Remove seed argument 

### LINKED ISSUE #415: Remove seed argument for `method_bayes`
Rstan defaults it's seed to `sample.int(.Machine$integer.max, 1)`. That is to say if the user has called `set.seed()` then Rstan will always have a fixed seed provided to it. As such the current seed argument to `method_bayes` doesn't really add value and provides confusing behaviour as it currently calls `set.seed()` internally which has side effects on the users session.

gowerc:
Closes #415

### DIFF
--- a/R/methods.R
+++ b/R/methods.R
@@ -92,12 +92,14 @@ method_bayes <- function(
     n_samples = 20,
     seed = NULL
 ) {
-    if (!is.null(seed)) {
-        warning(
-            "The `seed` argument to `method_bayes()` has been deprecated",
-            " please use `set.seed()` instead"
+    assertthat::assert_that(
+        is.null(seed),
+        msg = paste(
+            "The `seed` argument to `method_bayes()` has been deprecated;",
+            "please use `set.seed()` instead.",
+            collapse = " "
         )
-    }
+    )
 
     x <- list(
         burn_in = burn_in,
--- a/tests/testthat/test-mcmc.R
+++ b/tests/testthat/test-mcmc.R
@@ -604,7 +604,7 @@ test_that("fit_mcmc can recover known values with same_cov = FALSE", {
 
 
 test_that("seed argument to method_bayes is deprecated", {
-    expect_warning(
+    expect_error(
         {
             method <- method_bayes(
                 n_samples = 250,


data-raw/create_print_test_data.R
```diff
@@ -107,14 +107,14 @@ set.seed(413)
 dobj <- get_data(40)
 
 suppressWarnings({
+    set.seet(859)
```
gravesti, line 110:
```suggestion
    set.seed(859)
```

tests/testthat/test-mcmc.R
```diff
@@ -604,36 +603,17 @@ test_that("fit_mcmc can recover known values with same_cov = FALSE", {
...
-    expect_error(
-        draws(
-            dat,
-            dat_ice,
-            vars,
-            method_bayes(n_samples = 2, seed = NA),
-            quiet = TRUE
-        ),
-        regexp = "mcmc seed is invalid"
+test_that("seed argument to method_bayes is depreciated", {
```
gravesti, line 606:
```suggestion
test_that("seed argument to method_bayes is deprecated", {
```

R/methods.R
```diff
@@ -93,14 +90,20 @@ method_bayes <- function(
...
     same_cov = TRUE,
     n_samples = 20,
-    seed = sample.int(.Machine$integer.max, 1)
+    seed = NULL
 ) {
+    if (!is.null(seed)) {
+        warning(paste0(
+            "The `seed` argument to `method_bayes()` has been deprecated",
+            " please use `set.seed()` instead"
+        ))
```
gravesti, line 99:
```suggestion
        warning(
            "The `seed` argument to `method_bayes()` has been deprecated",
            " please use `set.seed()` instead"
        )
```
`paste0` is superfluous here, right?

gravesti (commented):
A few small comments :) @gowerc

gravesti (approved):
looks good @gowerc

gravesti:
Yeah, it's a bit in the middle now. One way would have been to throw a warning but still set the seed. However now that you're removed the functionality, I agree it might be better to go straight to error, since not setting the seed properly in analysis code will create worse problems.

gravesti (approved):

PR #435, opened 2024-10-03
Parallelisation support for analyse

### LINKED ISSUE #370: Possible to parallelize this?
https://github.com/insightsengineering/rbmi/blob/6769b203b000a453154f3a2a4295a870d9f1dfce/R/analyse.R#L155

Am passing a custom function here that is kind of (but not super) slow.  It would be nice to speed it up though.  Possible to change this to `mclapply`?

gowerc:
Closes #370 

Some general notes:

- I gave up on `future()` in the end. I just ran into edge case after edge case that made it too difficult to use and test. 
- I opted for just using parallel but allowing uses to provide their own clusters which can have their functions / libraries already loaded
- I then created a helper function `make_rbmi_cluster()` to make this setup process as smooth as possible
- This also allows for the re-use of clusters between runs to avoid having to repeatedly setup and tear down processes which is very expensive
- Unfortunately the performance gain was pretty underwhelming, I'm sure it will be better with analysis functions that are more expensive than `lm` but currently I am only seeing a 30% gain when using large samples e.g. >2000 for less than that it often takes longer to run in parallel. This is mostly due to the IO exchange by the looks of it. R's PSOCK clusters just transfer data exceedingly slowly :(

### DIFF
--- a/R/analyse.R
+++ b/R/analyse.R
@@ -221,12 +221,14 @@ analyse <- function(
     }
 
     # Mangle name to avoid any conflicts with user defined objects if running in a cluster
+    ..rbmi..analysis..imputations <- imputations
+    ..rbmi..analysis..delta <- delta
+    ..rbmi..analysis..fun <- fun
     objects <- list(
-        ..rbmi..analysis..imputations = imputations,
-        ..rbmi..analysis..delta = delta,
-        ..rbmi..analysis..fun = fun
+        ..rbmi..analysis..imputations = ..rbmi..analysis..imputations,
+        ..rbmi..analysis..delta = ..rbmi..analysis..delta,
+        ..rbmi..analysis..fun = ..rbmi..analysis..fun
     )
-    list2env(objects, envir = environment())
 
     cl <- make_rbmi_cluster(ncores)
 


R/parallel.R
```diff
@@ -1,41 +1,102 @@
...
 
-    cl <- parallel::makePSOCKcluster(
-        ncores
+    # Load user defined packages
+    packages <- if (is.null(packages)) {
+        # TODO - can't remember why this is needed; need to look into
+        "assertthat"
+    } else {
+        c(packages, "assertthat")
+    }
```
gravesti, line 61: This would be the same I think if packages is NULL or a character vector
```suggestion
    packages <- c(packages, "assertthat")
```

R/parallel.R
```diff
@@ -1,41 +1,102 @@
...
+    # Remove attempts to load rbmi as this will be covered later
+    packages <- grep("^rbmi$", packages, value = TRUE, invert = TRUE)
```
gravesti, line 63:
```suggestion
    packages <- setdiff(packages, "rbmi")
```

gravesti:
@gowerc Will you merge the lintr branch first? Do you still need to update the docs in this PR?

gravesti (approved):
Thanks for all the effort on this one @gowerc !

PR #437, opened 2024-10-04
Add support for different covariance structures

### LINKED ISSUE #385: Implement missing mmrm covariance structures
mmrm currently supports more covariance structures than what we actually expose. It would be good to enable all of these. Ideally it would be good if we could implement this as a transparent pass through so that we don't have to update the code each time they release a new one. 

(full list is available [here](https://openpharma.github.io/mmrm/main/reference/covariance_types.html) with more details available [here](https://openpharma.github.io/mmrm/main/articles/covariance.html))

This issue was originally raise in #384

gowerc:
Closes #385


gravesti (approved):
Nice that this can be expanded so easily @gowerc

PR #441, opened 2024-10-07
Convert rstan to suggest

### LINKED ISSUE #400: Consider making Stan a suggested package
It was noted in feedback that for some windows users installing stan was potentially problematic (assuming due to issues with rtools) however stan is only a minor optional subset of the code thus it might be nice to consider moving stan to being a suggested package and only compiling the stan model if it is used. 

Need to investigate if there would be any complications with regards to storing the compiled model. Ideally we'd want to store it in the package installation directory however I'm not sure if this would lead to permissions issues or violate any CRAN policies. Needs investigation.

gowerc:
Closes #400

### DIFF
--- a/.github/actions/build-src/action.yml
+++ b/.github/actions/build-src/action.yml
@@ -9,12 +9,22 @@ runs:
   using: "composite"
   steps:
 
+    - name: Determine system/package state
+      run: |
+        sink(".github/meta.txt")
+            list(
+                version = version,
+                pkgs = installed.packages()[c("rstan"), c("Version", "Package")]
+            )
+        sink()
+      shell: Rscript {0}
+
     - name: Cache Compiled Stan Code
       id: cache-pkgs
       uses: actions/cache@v3
       with:
         path: local/*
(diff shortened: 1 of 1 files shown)


R/utilities.R
```diff
@@ -515,18 +515,38 @@ as_dataframe <- function(x) {
...
+    local_file <- file.path("inst", "stan", "MMRM.stan")
+    system_file <- system.file(file.path("stan", "MMRM.stan"), package = "rbmi")
+    file_loc <- if (file.exists(local_file)) {
+        local_file
+    } else if (file.exists(system_file)) {
+        system_file
+    } else {
+        stop("Unable to find MMRM.stan; Please report this as a bug")
+    }
+    rstan::stan_model(
```
gravesti, line 547: Why not copy the file to `tools::R_user_dir("rbmi", which = "cache")` at this stage?

gravesti:
@gowerc I think you can't do much more than that wrt to STAN and compiler

PR #459, opened 2024-12-03
458 cran crash

gravesti:
I think this should solve the issue.

- Creates the model file with a hash containing the R version and versions of a number of important packages
- Deletes any other cached rbmi MMRM models which don't match the hash to prevent a build up


gravesti:
I have read the CLA Document and I hereby sign the CLA

gravesti:
recheck

PR #468, opened 2024-12-17
Update rbmi cache documentation

### LINKED ISSUE #466: Default cache path for rstan model causes rbmi to crash when run by the same user in parallel / inconsistent implement and comment
**Describe the bug**
The default cache path for rstan model causes rbmi to crash when run by the same user in parallel. This is caused by rstan recompiling the model each time (and race conditions occurring when this is done in parallel).

**To Reproduce**
Run the rbmi vignette on several cores using the same user account in parallel. 
I saw that in rbmi 1.3.1 the model file is now using a hash, which should also help with this (see #459)

**Environment (please complete the following information):**
 - OS: several
 - R version 4.4.1
 - rbmi version 1.3.0

**Proposed solution**
My proposal to fix this permanently is to change the default in https://github.com/insightsengineering/rbmi/blob/1d150e601a8191de4fa632599f4ededc7ad176d1/R/utilities.R#L620 to reflect the comment  

> #' Directory to store compiled Stan model in. If not set, a temporary directory is used for
> #' the given R session. Can also be set via the environment variable `RBMI_CACHE_DIR`.

That is, to set the default to `tempdir(check = TRUE)`, which is the _per-session temporary directory_ instead of the current default `tools::R_user_dir("rbmi", which = "cache")`, which is a _persistent_ per-user cache directory.

The workaround for the current version is to set `options("rbmi.cache_dir" = tempdir(check = TRUE))`.

Tagging @bailliem for awareness

gowerc:
Closes #466

### DIFF
--- a/R/utilities.R
+++ b/R/utilities.R
@@ -574,12 +574,12 @@ get_stan_model <- function() {
     cache_dir <- getOption("rbmi.cache_dir")
     dir.create(cache_dir, showWarnings = FALSE, recursive = TRUE)
     model_file <- file.path(cache_dir, paste0("MMRM_", get_session_hash(), ".stan"))
-    
+
     if (!file.exists(model_file)) {
         clear_model_cache()
         file.copy(file_loc, model_file, overwrite = TRUE)
     }
-    
+
     rstan::stan_model(
         file = model_file,
         auto_write = TRUE,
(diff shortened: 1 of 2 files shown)


luwidmer:
Thanks, this is much clearer than before :) !

gravesti (approved):

PR #477, opened 2025-02-05
Add `control_bayes()` and update code and documentation

danielinteractive:
This implements https://github.com/insightsengineering/rbmi/blob/main/misc/design_mcmc_improve.qmd.

I would like to note a few minor changes relative to the design:
- The `control` list is just a simple flat list, without an `additional` element.
- Additional helper function to complete the control list based on additional arguments inside the draws method
- Print method has been adapted


danielinteractive:
Thanks @gowerc for the review 😃

danielinteractive:
Thank you @gowerc , much appreciated. I adapted the code and docs accordingly, please have a look:

- print now works nicely also for more complex control elements
- seed setting has been more explained
- multiple chains are accounted for, the `n_samples` are distributed across the chains, and we account for the case that we need to generate slightly more samples first and then subset to the first `n_samples`

PR #484, opened 2025-02-18
Fix CRAN errors

### LINKED ISSUE #481: CRAN error libunwind.so.1
New CRAN error on their r-devel-linux-x86_64-fedora-gcc machine:

```
  ══ Failed tests ════════════════════════════════════════════════════════════════
  ── Error ('test-draws.R:144:9'): bayes ─────────────────────────────────────────
  Error in `fit_mcmc(designmat = model_df_scaled[, -1, drop = FALSE], outcome = model_df_scaled[, 
      1, drop = TRUE], group = data2[[vars$group]], visit = data2[[vars$visit]], 
      subjid = data2[[vars$subjid]], method = method, quiet = quiet)`: unable to load shared object '/tmp/RtmpBEQrRR/working_dir/RtmpXQXMbb/file11f85b3c2c931e.so':
    libunwind.so.1: cannot open shared object file: No such file or directory
  Backtrace:
      ▆
   1. ├─base::suppressWarnings(...) at test-draws.R:143:5
   2. │ └─base::withCallingHandlers(...)
   3. ├─rbmi::draws(d$dat, d$dat_ice, d$vars, meth, quiet = TRUE) at test-draws.R:144:9
   4. └─rbmi:::draws.bayes(d$dat, d$dat_ice, d$vars, meth, quiet = TRUE)
   5.   └─rbmi:::fit_mcmc(...)
```

https://www.r-project.org/nosvn/R.check/r-devel-linux-x86_64-fedora-gcc/rbmi-00check.html

I really struggle to believe this isn't a CRAN issue, my best guess is some form of built out of sequence error with something changing after rstan was compiled or that they are somehow persisting the working directory where we have cached the models so that they aren't being cleared properly...

Must admit I am getting a bit tired of these constant stan breakages :(

A couple of options we could consider though all have tradeoffs:

1) Switch out rstan for cmdstanr. Means we have to disable lots of tests on CRAN and means we have to deal with internal Roche issues of getting cmdstanr available. Doesn't necessarily fix the issue if the issue is due to the model caching and not the binaries. 

2) Simply disable the rstan tests on CRAN. We can't replicate these errors anywhere else other than a single CRAN machine which runs non-standard software versions, it feels very unlikely these issues are ones that wouldn't come up in regular usage. Potentially risks fallout with the CRAN maintainers if they realise our solution was to just disable the tests and may potentially mask future compatibility issues if an interface with rstan does change. 

3) A variant on (2) we could just disable any sampling on CRAN but then store pre-sampled values. Essentially means we don't have to disable quite as many tests as we would do with (2) but still means we mask other potentially future issues and bloat the repo with test-only data. 

4) Disable the cache entirely. There is a reasonable chance the issue is due to the cache so this might side step it but means we force people to have to recompile the models each time which is quite lengthy with rstan.  Alternatively the cache could just be disabled by default with an option to enable it for users that know they need it ? This may also not fix the issues if its due to underlying software versions / rstan / cran issues. 

Any thoughts / suggestions  @gravesti  / @danielinteractive  ?

gowerc:
Closes #481

Issue is due to a `#` character in our comments which the new C23 parser interprets as a pre-processor command. This should resolve that. 

I also added a `rbmi_` prefix to help avoid any potential name conflicts in the R session directory (currently we delete any file starting with "MMRM_"

I also attempted to implement @gravesti 's idea of conditionally disabling the cache entirely if run on CRAN. Though I implemented this essentially as disable cache if `R CMD CHECK` is run and `NOT_CRAN != TRUE`.

### DIFF
--- a/R/utilities.R
+++ b/R/utilities.R
@@ -689,7 +689,7 @@ get_stan_model <- function() {
 set_options <- function() {
 
     cache_dir <- Sys.getenv("RBMI_CACHE_DIR", unset = tools::R_user_dir("rbmi", which = "cache"))
-    enable_cache <- Sys.getenv("RBMI_ENABLE_CACHE", unset = "TRUE") == "TRUE"
+    enable_cache <- isTRUE(as.logical(Sys.getenv("RBMI_ENABLE_CACHE", unset = "TRUE")))
 
     current_opts <- names(options())
     rbmi_opts <- list(


R/utilities.R
```diff
@@ -680,15 +688,13 @@ get_stan_model <- function() {
 #' @name rbmi-settings
 set_options <- function() {
 
-    cache_dir <- Sys.getenv("RBMI_CACHE_DIR")
-
-    if (cache_dir == "" || is.null(cache_dir)) {
-        cache_dir <- tools::R_user_dir("rbmi", which = "cache")
-    }
+    cache_dir <- Sys.getenv("RBMI_CACHE_DIR", unset = tools::R_user_dir("rbmi", which = "cache"))
+    enable_cache <- Sys.getenv("RBMI_ENABLE_CACHE", unset = "TRUE") == "TRUE"
```
gravesti, line 692:
```suggestion
    enable_cache <- isTRUE(as.logical(Sys.getenv("RBMI_ENABLE_CACHE", unset = "TRUE")))
```
maybe use as.logical to catch `c("T", "TRUE", "True", "true")`

gravesti (approved):
Looks good to me @gowerc 
Just one suggestion

gravesti (approved):
Thanks @gowerc

PR #495, opened 2025-05-16
Add MCSE

### LINKED ISSUE #493: Monte Carlo Standard Errors for p-values
**Is your feature request related to a problem? Please describe.**

The Bayesian approach to multiple imputation yields probabilistic results. The precision of the MCMC algorithm can be quantified through the MC standard error (MCSE). Such quantification is important for judging the scientific reproducibility of the results, e.g., when computational parameters such as the seed are changed. For the treatment effect estimates obtained through Rubin's rule, a closed-form expression for the MCSE exists: MCSE = $\sqrt{SE/N_{imp}}$. However, since decisions are generally based not just on the treatment effect estimate but also on the p-value, it is important to calculate the corresponding MCSE. No closed-form expression exists for the MCSE of the p-value, but as argued by [Royston et al](https://journals.sagepub.com/doi/pdf/10.1177/1536867X0900900205), the Jackknife method could be used to estimate the MCSE.

**Describe the solution you'd like**
I see two approaches for incorporating this feature into _rbmi_ that might be of interest:
- Function to be applied to the analysis object that calculates the MCSE of the pooled estimates.
- As part of pool, calculate the MCSE of the pooled estimates and store them in the pool object.

**Describe alternatives you've considered**
I've implemented the MCSE calculation for the p-value on an ad-hoc basis for the example from the _Quickstart_ vignette. Please see the code below. 

```r
library(rbmi)
library(tidyverse)

data("antidepressant_data")
dat <- antidepressant_data

# Use expand_locf to add rows corresponding to visits with missing outcomes to the dataset
dat <- expand_locf(
  dat,
  PATIENT = levels(dat$PATIENT), # expand by PATIENT and VISIT 
  VISIT = levels(dat$VISIT),
  vars = c("BASVAL", "THERAPY"), # fill with LOCF BASVAL and THERAPY
  group = c("PATIENT"),
  order = c("PATIENT", "VISIT")
)

# Create data_ice and set the imputation strategy to JR for
# each patient with at least one missing observation
dat_ice <- dat %>% 
  arrange(PATIENT, VISIT) %>% 
  filter(is.na(CHANGE)) %>% 
  group_by(PATIENT) %>% 
  slice(1) %>%
  ungroup() %>% 
  select(PATIENT, VISIT) %>% 
  mutate(strategy = "JR")

# In this dataset, subject 3618 has an intermittent missing values which does not correspond
# to a study drug discontinuation. We therefore remove this subject from `dat_ice`. 
# (In the later imputation step, it will automatically be imputed under the default MAR assumption.)
dat_ice <- dat_ice[-which(dat_ice$PATIENT == 3618),]

# Define the names of key variables in our dataset using `set_vars()`
# and the covariates included in the imputation model
# Note that the covariates argument can also include interaction terms
vars <- set_vars(
  outcome = "CHANGE",
  visit = "VISIT",
  subjid = "PATIENT",
  group = "THERAPY",
  covariates = c("BASVAL*VISIT", "THERAPY*VISIT")
)

# Define which imputation method to use 
method <- method_bayes(n_samples = 250)

# Create samples for the imputation parameters by running the draws() function
set.seed(987)
drawObj <- draws(
  data = dat,
  data_ice = dat_ice,
  vars = vars,
  method = method,
  quiet = FALSE
)

# Impute the data
imputeObj <- impute(
  drawObj,
  references = c("DRUG" = "PLACEBO", "PLACEBO" = "PLACEBO")
)

# Fit the analysis model on each imputed dataset
anaObj <- analyse(
  imputeObj,
  ancova,
  vars = set_vars(
    subjid = "PATIENT",
    outcome = "CHANGE",
    visit = "VISIT",
    group = "THERAPY",
    covariates = c("BASVAL")
  )
)

# Pool the results
poolObj <- pool(
  anaObj, 
  conf.level = 0.95, 
  alternative = "two.sided"
)
poolObj



##############################################
# MCSE of p-value
##############################################

# Step 1: Calculate the leave-one-out p-values
# Idea: modify the analysis object and remove the results from the i-th imputed dataset
jackknifePvalue <- lapply(seq_along(anaObj$results), 
                          FUN = function(i){
                            x <- anaObj
                            # Remove i-th results
                            x$results <- x$results[-i]
                            # Recreate class attribute for results
                            attr(x$results,"class") <- c("rubin", "list")
                            # Adjust the samplesize to match the length of the results object
                            x$method$n_samples <- x$method$n_samples - 1
                            
                            # Run Rubin's rule on modified analysis object
                            poolObjX <- pool(x, conf.level = 0.95, alternative = "two.sided")
                            
                            # Create tibble with all p-values
                            tibble(param = names(poolObjX$pars),
                                   pvalue = map_dbl(poolObjX$pars, ~.x$pvalue))
                            
                          })

# Step 2: Calculate MCSE of p-value for each parameter
bind_rows(jackknifePvalue, .id = "Imp") |> 
  group_by(param) |> 
  summarize(pvalueVar = var(pvalue),
            mcse = sqrt((n()-1)^2*pvalueVar / n()))
```

danielinteractive:
closes #493

Notes:

- Uses Jackknife for all parameters for simplicity
- Uses same structure for `mcse()` result as `pool()` to make it easy for the user
- Describes the method in the statistical specifications vignette and the `pool()` documentation


danielinteractive:
Hi @gowerc and @tobiasmuetze, this is ready for review from your side, thank you 😄

vignettes/stat_specs.Rmd
tobiasmuetze, line 1: Could we also reference Efron and Gong (https://doi.org/10.2307/2685844) who first discussed using Jackknife to calculate the SE of a parameter?

tests/testthat/_snaps/pool.md
tobiasmuetze, line 1: - I am wondering if it would be better to not have `mcse()` being discussed as part of the manual for `pool()`. This would allow more space to also explain the function in more detail.

tobiasmuetze, line 1: Shouldn't the output also include some details of the pool-Object such as that the confidence level for the CIs is 0.95?

DESCRIPTION
tobiasmuetze, line 1: Update Craig's email address?

R/pool.R
```diff
@@ -29,6 +27,9 @@
 #' - `method_approxbayes()` & `method_bayes()` both use Rubin's rules to pool estimates
 #'  and variances across multiple imputed datasets, and the Barnard-Rubin rule to pool
 #'  degree's of freedom; see Little & Rubin (2002).
+#'  Here, the `mcse()` function can compute the Monte Carlo standard error (MCSE) of the
```
tobiasmuetze, line 30: add reference to Efron and Gong (https://doi.org/10.2307/2685844)

tests/testthat/_snaps/pool.md
danielinteractive: I thought about that too. I decided then to keep them together, because `mcse()` is basically an "add-on" for the `pool()` functionality. Also because the output is "parallel" to each other, and the `mcse()` use case is relatively small (only applies to the Bayesian flavors)

danielinteractive: Since the functions are used together, I think that it is more elegant to not repeat information in the `mcse()` output. Also, it is not really needed to know the confidence level of the CIs - the MCSE is orthogonal to that I would say. The other reason is that like this the functionality is easier extensible - e.g. if a user adds their own statistics to the `pool()` object (via a fork).

DESCRIPTION
danielinteractive: will do

R/pool.R
danielinteractive: yep

tests/testthat/_snaps/pool.md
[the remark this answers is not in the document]
danielinteractive: good point, will change

tobiasmuetze (approved):

PR #514, opened 2025-07-15
Update docker image

gowerc:
Adding in jinjar package (and a few others) so that Daniels PR unit tests can actually run.


danielinteractive (approved):
Thanks @gowerc !

PR #515, opened 2025-07-16
Update Docker Pipeline

gowerc:
This PR attempts to address some short comings in the current docker build process namely:

- This enables the build to happen automatically when anything edits the docker folder (currently only runs on manual trigger or on bi-weekly schedule)
- Updated so that it will only push to "latest" if the push-event was on the main branch
- Removed the GHA caching code, I don't think this actually adds any value as we install binary packages now which is very quick and the article that was linked to about caching no longer exists, instead I've just included the defaults as recommend [here](https://docs.docker.com/build/ci/github-actions/cache/)


gravesti:
@gowerc This all looks reasonable to me, but I don't think I'm enough of a GHA expert to know that it is 100% correct. Should we merge and see?

gravesti (approved):

PR #520, opened 2025-08-18
Support for multiarm ancova

### LINKED ISSUE #145: Ancova to support > 2 arms

gowerc:
Closes #145 

Hey @danielinteractive  /  @tobiasmuetze,

This contains my initial code proposal for implementing multiarm ancova support (note that it is currently missing documentation and tests).

Some key points:

- To preserve backwards compatibility I've left the original `ancova` function unchanged. To support multiple groups a new naming scheme is required for the parameters which would break any existing code which is looking for the current names like `trt` / `lsm_rf`.

- That is, if users want to use the new multi-group ancova they will have to explicitly set `fun=ancova_m_groups` in the call to `analyse()`

- For the new naming convention I've just used the level numbers e.g.   `lsm_L0` = least squared means for the reference, `lsm_L1` = least squared means for the first offset / coefficient.  I've also used a more explicit `trt_L1_L0` to mean the coefficient of the first offset from the reference.   

- As I think we agreed I'm currently only extracting the treatment effects against the reference e.g.  `L1 - L0` & `L2 - L0`. I'm not calculating / extracting all pairwise combinations. 

- For the naming convention I didn't want to use the user-provided group/level values as that can lead to a ton of special edge cases with whitespaces / special characters that are hard to predict and make value-extraction very error prone. Instead I opted for just renaming the group variable to `rbmiGroup` and re-leveling it to be `L0, L1, etc`. Only complication here was that the formula provided in theory can have interaction terms so had to create a function to recursively traverse the AST updating any references of `group` to `rbmiGroup`. I will need to put in extensive unit testing to make sure theres no edge cases here with things like `I()` functions. 

If there any no main objections here I'll start trying to add docs and tests.


danielinteractive:
> Interestingly despite what their documentation says I don't think they actually support more than 2 arms, in particular they have near the top of the function `checkmate::assert_factor(data[[group]], n.levels = 2L)` which would error if there were more than 2 arms.

Hmm... ok weird. But I am pretty sure we wanted this to work with multiple arms, that is also what the other code suggests... I will need to check in September when I am back.

> I am very conflicted here, I agree with everything you said and I am also not happy with splitting it out but that being said I don't see an easy way of merging them without breaking backwards compatibility as the naming scheme has to be different in order for it to make sense we currently have:
> 
> ```
> lsm_alt_visit_1
> lsm_ref_visit_1
> trt_visit_1
> ```

Yeah I understand the need for backwards compatibility. But I wonder if an alternative solution here could be to have this old naming scheme used for 2 arms, and the new naming scheme used for more than 2 arms?

> In hindsight these are already bad names that confuse users. If we were to extend this whilst maintaining backwards compatibility we'd end up with something like:
> 
> ```
> lsm_alt_visit_1
> lsm_alt2_visit_1
> lsm_ref_visit_1
> trt_visit_1
> trt_alt2_visit_1
> ```
> 
> Which I would argue is even worse.

Personally I think that would also be acceptable. The user will not see these names too much anyway if they just use the `rbmi` functions.

> 
> Alternatively we could just have some code that dispatches to the different ANCOVA function based on the number of levels in group variable but this makes the documentation / explanation clunky e.g.

Yeah I would not have two functions necessarily but just two naming schemes. Or just go with the naming scheme just described which I think is fine.

> One option though to minimise maintenance is that we could deprecate `anocva` with a note that it will be removed in the future and replaced with `ancova_m_groups` then in 6 months to a years time we could fully remove it, we could also create an alias of `ancova = ancova_m_groups` after removing so we don't have such an awkward name.
> 
> (Despite my tone I am not confident on the best path forward here as all options appear bad to me so please do challenge if you still disagree)

Personally I would just go with the naming scheme you suggested, I think that is reasonable. Just for the sake of a better naming scheme I would not go the route of two functions or function deprecation. Just my 2 cents 😄

wwojciech:
Hi @gowerc @tobiasmuetze @danielinteractive,

This functionality would be really helpful for the project I'm currently working on.

Is there any chance this PR might get merged?

Thanks!
Wojtek

danielinteractive:
@wwojciech I think we were not yet fully aligned on the right API here for the function, and @gowerc is still on leave at the moment... @luwidmer what is your call on this one from maintainer perspective?

wwojciech:
Hi @danielinteractive @luwidmer,

To make downstream parsing easier and more robust, have you considered exposing the components currently concatenated into the `parameter` column as separate columns in the output of `analyse()` -> `pool()`?

For example:

 ```
   estimate_type  group_var group_level_1 group_level_2 visit    est
   <chr>          <chr>     <chr>         <chr>         <chr>    <dbl> 
1 lsm            TRT01P    Placebo       NA            Week16   55.8
2 lsm            TRT01P    A             NA            Week16   64.2
3 lsm            TRT01P    B             NA            Week16   66.7   ......
4 contrast       TRT01P    A             Placebo       Week16   8.4
5 contrast       TRT01P    B             Placebo       Week16   10.9 
 ```
 for `var$group` having 3 levels: `"Placebo"`, `"A"`, `"B"` (in this order).

wwojciech:
Hi @gowerc , @luwidmer - can you please update on this PR?

luwidmer:
@tobiasmuetze, from your experience implementing this ad-hoc, what are your thoughts on what this should look like in rbmi?

tobiasmuetze:
I am generally fine with the idea that `ref` refers to the first factor level of `vars$group`, `alt1` to the second factor level, etc. However, I think the output of `pool()` should be more explicit about the underlying study arms and the corresponding comparisons. In that respect, I agree with @wwojciech’s proposal, but I would suggest extending the `pool()` output while retaining the current `parameter` variable for backwards compatibility.

I would also suggest allowing users to specify the contrasts of interest, since there may be cases where two `alt` levels need to be compared directly. The default could remain all comparisons of each `alt` level against the `ref` level.

wwojciech:
Thanks, @tobiasmuetze.

I have a positive feeling about keeping `parameter` column for backward compatibility. In that case, the `parameter` column in the `pool()` output could look like this:
```
parameter
<chr>
lsm_L1_Week16
lsm_L2_Week16
lsm_L3_Week16
contr_L1_L3_Week16
contr_L2_L3_Week16
contr_L2_L1_Week16
```
as a result of the following settings:

1. `var$group` has 3 levels: `"A"`, `"B"`, and `"Placebo"` (in this order).
2. The user requested the following contrasts: 

- `"A"` - `"Placebo"`,
- `"B"` - `"Placebo"`,
- `"B"` - `"A"`.

One additional suggestion: if possible, I would propose numbering the levels from `L1` rather than `L0`. This would be more consistent with R conventions, where (for example, in `lm()`) the reference level is simply the first factor level rather than "level 0." I think starting with `L1` would therefore be more intuitive for R users.

tobiasmuetze:
> One additional suggestion: if possible, I would propose numbering the levels from L1 rather than L0. This would be more consistent with R conventions, where (for example, in lm()) the reference level is simply the first factor level rather than "level 0." I think starting with L1 would therefore be more intuitive for R users.

I would agree in principle if we were to start building this from scratch. However, giving that `{rbmi}` is used in a regulated environment for decision making, I'd prefer to prioritize backward compatibility in this case and the usage of ref, alt, alt2, etc., and keeping the parameter structure the same. 

Any suggestions how to specify the contrast would be appreciated.

wwojciech:
> > One additional suggestion: if possible, I would propose numbering the levels from L1 rather than L0. This would be more consistent with R conventions, where (for example, in lm()) the reference level is simply the first factor level rather than "level 0." I think starting with L1 would therefore be more intuitive for R users.
> 
> I would agree in principle if we were to start building this from scratch. However, giving that `{rbmi}` is used in a regulated environment for decision making, I'd prefer to prioritize backward compatibility in this case and the usage of ref, alt, alt2, etc., and keeping the parameter structure the same.
> 
> Any suggestions how to specify the contrast would be appreciated.

Sure, thank you @tobiasmuetze. 

Then, there must either be a clear definition of the relationship between `ref`, `alt`, `et2`, etc., and the levels of `group`, or the output table of `pool()` should follow the structure I proposed above (of course, with the `parameter` column retained for backward compatibility). Otherwise, `group`-level (label) identification may become error-sensitive.

### Constrast specification

One approach that comes to mind naturally would be to specify contrasts through an additional `group_contrasts` argument in `set_vars()`. This could be a list of pairs of levels of `group`, where the first element in each pair is the minuend. For example:

```r
group_contrasts = list(
  c("A", "Placebo"),
  c("B", "Placebo")
)
```

This would represent the contrasts `A - Placebo` and `B - Placebo`, respectively.

Alternatively, this could be represented as a `matrix` or a named `list`, if there is an opportunity for `{rbmi}` to make use of the names. For example, names could be propagated to outputs such as the table produced by `pool()`.

I proposed the `group_` prefix so that the argument is conceptually connected to the `group` argument of `set_vars()`.

tobiasmuetze:
> Then, there must either be a clear definition of the relationship between ref, alt, et2, etc., and the levels of group, or the output table of pool() should follow the structure I proposed above (of course, with the parameter column retained for backward compatibility). Otherwise, group-level (label) identification may become error-sensitive.

Agree!

> Constrast specification

At first sight, I am fine with the `group_contrasts` proposal, but need to think the implications fully through.

R/ancova.R
```diff
@@ -206,3 +322,110 @@ ancova_single <- function(
...
+    outcome,
+    group,
+    covariates,
+    weights = c("counterfactual", "equal", "proportional_em", "proportional")
+) {
+    weights <- match.arg(weights)
+    assert_that(
+        is.factor(data[[group]]),
+        length(levels(data[[group]])) >= 2,
+        length(unique(data[[group]])) == length(levels(data[[group]])),
```
wwojciech, line 385: @tobiasmuetze - Can we please remove this from the assertion:
```r
length(unique(data[[group]])) == length(levels(data[[group]]))
```

and then, of course, modify the code below that computes `trt` to:

```r
trt <- lapply(
  levels(data2[["rbmiGroup"]])[-1],
  function(x) {
    term <- sprintf("rbmiGroup%s", x)
    if (term %in% names(coef(mod))) {
      list(
        est = coef(mod)[[term]],
        se = sqrt(vcov(mod)[term, term]),
        df = df.residual(mod)
      )
    } else {
      list(est = NA, se = NA, df = NA)
    }
  }
)

```
This is because there are some cases where we have no records for a certain level of `data[[vars$group]]`, and this should not cause the function to stop.

wwojciech:
Also, @tobiasmuetze , please revisit the existing implementation of `ancova_single_m_group()`. The current code implicitly assumes that treatment (`contr.treatment()`) contrasts are used for the grouping factor. If a different contrast coding is active (e.g. `contr.sum()` or another custom contrast), the coefficient extraction no longer corresponds to treatment-vs-reference comparisons, which could lead to incorrect or misleading results without necessarily producing an error.

R/ancova.R
tobiasmuetze: Is my understanding correct that if this assertion is removed, the group variable in the input data set could have a certain level but no subjects with this level? In which clinical trial scenario would this be relevant?

tobiasmuetze:
> Also, @tobiasmuetze , please revisit the existing implementation of ancova_single_m_group(). The current code implicitly assumes that treatment (contr.treatment()) contrasts are used for the grouping factor. If a different contrast coding is active (e.g. contr.sum() or another custom contrast), the coefficient extraction no longer corresponds to treatment-vs-reference comparisons, which could lead to incorrect or misleading results without necessarily producing an error.

I agree that the current coefficient extraction only supports the originally discussed setup, where each group level is compared against the first/reference level. However, based on our recent discussion, we now want to support user-specified contrasts, rather than assuming comparisons to the first level.

R/ancova.R
wwojciech: > Is my understanding correct that if this assertion is removed, the group variable in the input data set could have a certain level but no subjects with this level? In which clinical trial scenario would this be relevant?

Yes, it is. There might be no data for a given level of `data[[vars$group]]`.

This is relevant in scenarios where we fit an ANCOVA model using a subset of the full dataset defined by a subgroup variable (such as geographic region). In that case, for some levels of the subgroup variable, certain levels of `data[[vars$group]]` may be absent from the subset.

For example, in a given geographic region, we may not have data for all treatment groups. See the example table below for a dataset with three treatment groups (i.e., `data[[vars$group]]` has three levels), where the table presents results for only two treatment groups of interest.

example_table

tobiasmuetze: I’d suggest keeping the assertion, because this scenario seems to point to an inadequate model/data setup rather than something the function should accommodate.

For subgroup analyses based on multiply imputed data, the treatment-by-subgroup interaction should be included in the imputation model. If a treatment-by-subgroup combination is completely absent in the data, then fitting such an imputation model should already be problematic. 

In that sense, the assertion `length(unique(data[[group]])) == length(levels(data[[group]]))` at least ensures that all declared treatment levels are actually represented in the dataset being analyzed. If a level is absent after subsetting, that should probably be treated as a signal that the analysis model is not adequate for the intended subgroup analysis.

The statistical concern is that if the imputation model does not include the treatment-by-subgroup interaction, subsequent subgroup-specific treatment effect estimates, or analyses of treatment-by-subgroup interaction, may be biased toward the null when interactions are present; see [Schafer](https://journals.sagepub.com/doi/10.1177/096228029900800102). More generally, [Austin et al.](https://doi.org/10.1016/j.cjca.2020.11.010) state that “it is important to include in the imputation model all the variables that will be included in the analysis model.”

wwojciech: > I’d suggest keeping the assertion, because this scenario seems to point to an inadequate model/data setup rather than something the function should accommodate.
> 
> For subgroup analyses based on multiply imputed data, the treatment-by-subgroup interaction should be included in the imputation model. If a treatment-by-subgroup combination is completely absent in the data, then fitting such an imputation model should already be problematic.
> 
> In that sense, the assertion `length(unique(data[[group]])) == length(levels(data[[group]]))` at least ensures that all declared treatment levels are actually represented in the dataset being analyzed. If a level is absent after subsetting, that should probably be treated as a signal that the analysis model is not adequate for the intended subgroup analysis.
> 
> The statistical concern is that if the imputation model does not include the treatment-by-subgroup interaction, subsequent subgroup-specific treatment effect estimates, or analyses of treatment-by-subgroup interaction, may be biased toward the null when interactions are present; see [Schafer](https://journals.sagepub.com/doi/10.1177/096228029900800102). More generally, [Austin et al.](https://doi.org/10.1016/j.cjca.2020.11.010) state that “it is important to include in the imputation model all the variables that will be included in the analysis model.”

Thank you @tobiasmuetze 

My understanding is as follows.

Assume that the imputation model was correctly specified (following the recommendations of Schafer and Austin), i.e., the treatment-by-subgroup interaction is included in the imputation model. Even in this case, it may still happen that, for certain levels of the subgroup variable, not all treatment groups are represented in the observed data. As a result, after imputation there will still be no observations for that particular treatment-by-subgroup combination, since multiple imputation imputes missing values for existing subjects rather than creating new observations.

However, I do not think this necessarily prevents a valid analysis within that subgroup. If the estimand of interest does not involve the missing treatment level, then contrasts among the remaining observed treatment groups remain estimable. In that situation, I would expect the ANCOVA model to fit successfully and to return valid estimates for the estimable contrasts, while returning `NA` for treatment effects involving the absent treatment level.

For this reason, I think the assertion may be too restrictive.

@luwidmer, @danielinteractive - what do you think?

luwidmer:
@tobiasmuetze & @wwojciech: any objections to merging into this branch the v2 version that implements user-specified contrasts (#590) for further discussion / iteration (if needed)?

wwojciech:
> @tobiasmuetze & @wwojciech: any objections to merging into this branch the v2 version that implements user-specified contrasts (#590) for further discussion / iteration (if needed)?

Hi @luwidmer - I have no objections. That said, I must admit I didn’t have a chance to check all the details and only had a very brief look at it.

R/ancova.R
luwidmer: What would you think of the following: relax the assertion, but not with a silent `NA` to address these two concerns:

* Imputation / analysis model alignment (@tobiasmuetze): the treatment-by-subgroup interaction must be in the imputation model, or subgroup effects bias toward the null (Schafer; Austin et al.). Agreed.
* Structural empty cells (@wwojciech): MI imputes outcomes for existing subjects, never creates new ones. So a congenial model can still leave an arm×subgroup cell empty by design (e.g. an arm that never opened in a region). 

However, then `length(unique(data[[group]])) == length(levels(data[[group]]))` is the wrong check to do?
It is neither necessary (estimable contrasts don't need every level) nor sufficient (one subject per level does not guarantee the treatment-by-subgroup interaction being in the imputation model) for the above criteria I think?
If the estimand excludes the absent arm, the remaining contrasts could still be identified?

But I'd keep @tobiasmuetze's signal instead of suppressing it, e.g., replace `stop()` with a `warning()` naming the empty level, return estimable contrasts + `NA` for the rest. That distinguishes "empty by design" from a data-prep bug, which a silent `NA` can't?

wwojciech: Thank you @ludmier. I agree with your points. Having said that, I would be slightly cautious about issuing a warning here. A user could repeatedly get warnings for perfectly valid analyses, making the output noisy.

Is there any way to be more specific, i.e. identify a data-preparation issue such as imputation model misspecification and issue the warning only in such cases?

----
Apart from data-preparation issues, I would also distinguish between the following two cases:

1. An empty treatment level, but the requested contrasts are still estimable.
    -> I would not warn. Return the estimable contrasts and NA for the non-estimable ones.
2. The user explicitly requests a contrast involving the absent treatment.
    -> This is more actionable. A warning would make sense, since the requested estimand cannot be estimated.

That said, I am not trying to push strongly for one solution here. Whether you decide to keep the warning or not is ultimately fine with me, as long as we don't `stop()` :)

munoztd0:
@luwidmer and @tobiasmuetze sorry to bother but any plans to merge it into main now that the new API has been merged in ?

danielinteractive:
I understand that we have an ooo situation so need to wait a couple of days - that said, if there is any support needed here happy to jump in any time from my side.

vignettes/quickstart.Rmd
```diff
@@ -229,6 +229,55 @@ function which determines the names of the key variables within the data and the
 
 Please also note that the names of the analysis estimates contain `ref` and `alt` to refer to the two treatment arms. In particular `ref` refers to the first factor level of `vars$group` which does not necessarily
 coincide with the control arm. In this example, since `levels(dat[[vars$group]]) = c("DRUG", PLACEBO`), the results associated with `ref` correspond to the intervention arm, while those associated with `alt` correspond to the control arm.
+<<<<<<< HEAD
```
danielinteractive: seems some conflict markers are still in here

vignettes/quickstart.Rmd
```diff
@@ -229,6 +229,55 @@ function which determines the names of the key variables within the data and the
...
+    group_contrasts = list(
+        pooled_vs_pbo = c(Placebo = -1, Low = 0.5, High = 0.5)
+    )
+)
+anaObj <- analyse(imputeObj, ancova, vars = vars)
+```
+
+See the documentation for `ancova()` and `set_vars()` for more details.
+=======
+>>>>>>> origin/main
```
danielinteractive: here too

NEWS.md
```diff
@@ -1,5 +1,23 @@
...
+  of treatment-group contrasts for `ancova()`. By default a contrast of each non-reference
+  group versus the reference group is estimated. Custom contrasts must be named (the name
+  becomes the output `parameter` name) and may be given as pairwise
+  `c(minuend, subtrahend)` character vectors or as general named numeric weight vectors
+  over the group levels (e.g. `c(Placebo = -1, A = 0.5, B = 0.5)` for a pooled comparison);
+  contrasts are evaluated in a contrast-coding-agnostic way. (#520)
+* The `data.frame` produced by `as.data.frame()` / `pool()` now includes explicit
+  `estimate_type`, `group`, `group_level_1`, `group_level_2`, `contrast_label` and `visit`
+  columns when the analysis was performed with `ancova()`. The existing `parameter` column
+  is retained for backwards compatibility. (#520)
```
danielinteractive, line 17: Maybe combine this into a single item talking about ancova for more than two groups?

danielinteractive (commented):
Thanks again @luwidmer for the nice proposal, I just had a quick look through and it looks really nice. Also the testing seems to be very thorough which is great to see, especially that we get full backwards compatibility. Minor things below.

luwidmer:
@gowerc could you take a look as to whether you're fine with what I did to your branch 😅 ?

tobiasmuetze:
I reviewed the functionality from a user perspective. I have no suggestions for changes. Great work and very useful.

gowerc:
AI review flagged 2 issues for me:

- (major) the LSMs were not invariant to contrasts as it looks like the code was just extracting the beta coeficient which only works for corner-point constraint.  Added a bunch more processing code to handle the different contrasts by constructing a small reference matrix and then using that to pick out / sum the relevant beta coeficients.

- (minor) there was an issue where you could get a parameter name clash if you had overlapping visit / treatment labels. I didn't bother fixing this as you have to be going out of your way to cause it so just put an assertion that looks for duplicate labels and then throws an error. 

[9f6c0b8](https://github.com/openpharma/rbmi/pull/520/commits/9f6c0b8a8719757b97734709b7befe5af2ef2048) @danielinteractive  / @luwidmer  if you have a chance to see what you think. 

Other wise I think the code looks good to me.

danielinteractive (approved):
Super, thanks @gowerc - I agree that the now general encoding support for the factors is a good idea.

gowerc:
On reflection think I've confused myself as to what exactly we want or expect the contrast to represent. Apologies if this was in the above conversation, I tried reading but I was struggling to follow :(

So in the current implementation on main we naively force the use corner-point-constraint  (`contr.treatment`) and then just report the treatment contrast effect as the corresponding regression Beta coeficient. In hindsight I'm not even sure if this is correct as this is implicitly equivalent to setting all covariate values + interactions to 0  e.g. ignoring interactions and taking the reference level for categorical covariates.  For example is it meaningful to have an age of 0 or assume everyone is Male ? (Edit - I appreciate if theres no treatment interaction terms then the implementation here is fine, but I don't think its right if there is a treatment interaction term).

The current implementation in this PR (as of my most recent commit) slightly addresses this by making the choice of covariates explicit (internally at least) and allowing for non-corner-point-constraint codings. It does this by building a reference dataset for each treatment level where continuous covariates are set to 0 and categorical covariates are set to the reference level using the contrast coding scheme  e.g.

corner-point
```
> mod_treatment <- lm(out ~ rbmiGroup * sex + I(age^2), data = dat)
> design_treatment <- ancova_contrast_design(mod_treatment, dat, "out")
> design_treatment
  (Intercept) rbmiGroupL2 sexMale I(age^2) rbmiGroupL2:sexMale
1           1           0       0        0                   0
2           1           1       0        0                   0
```

sum-to-zero
```
> mod_sum <- lm(out ~ rbmiGroup * sex + I(age^2), data = dat)
> design_sum <- ancova_contrast_design(mod_sum, dat, "out")
> design_sum
  (Intercept) rbmiGroup1 sex1 I(age^2) rbmiGroup1:sex1
1           1          1    1        0               1
2           1         -1    1        0              -1
```

This is then used by the user defined contrasts list to decide which rows / fitted values be calculated and then taking the difference between them. 

I'm just a bit lost now as to wether it still makes sense to be setting covariates to 0 / the reference level for categorical values?  Would it not be more reasonable to just do the difference of LSMs which then has weighting schemes to determine what values to set the covariates at which is representative ? 

@tobiasmuetze  / @danielinteractive  / @luwidmer  / @wwojciech

wwojciech:
Hi @gowerc 

I think these are very valid considerations. A bit aside from the technical discussion, I was wondering whether the current state of the PR would allow us to split this into a separate issue, so that the custom contrast could be considered/implemented in a dedicated PR, while this PR could be completed with the multi-arm support?

If this would not introduce too much mess or require significant additional work, I would be very much in favour of this option. The package I support with @danielinteractive currently requires the multi-arm support first, while, as far as I am aware, the custom contrast is less critical for us (if at all).

@luwidmer , @tobiasmuetze - for your consideration.

gowerc:
hmm, I mean the concern I raised is tangental to the custom contrast, I think the custom contrast code implemented here by @luwidmer  is working as intended. My issue is more around wether it is creating a meaningful metric.  Though I think this is an issue with the `main` version of rbmi as well.  Given the code is correct as per its intended purpose I am happy to merge this into `main` and then create a separate issue for the interactions conversation but I do think the interactions bit needs resolution before I would feel happy pushing to CRAN.

danielinteractive:
@gowerc yeah makes sense. Maybe @wwojciech you can investigate this concern/question with examples to check if it is a problem?

R/ancova.R
gowerc: Hey @danielinteractive  / @wwojciech  / @luwidmer  - Was this thread resolved, it being open is blocking me merging :)

gowerc:
I have moved the interaction conversation then to #595 ,  I am happy to merge but there appears to be a lingering unresolved conversation thread which is blocking the merge (I'd rather not admin force merge unless I really have to)

danielinteractive:
@gowerc I think it has been resolved now

R/ancova.R
gowerc: I'm going to resolve this thread so I can merge the PR. At the moment the assertion appears to still be in place:

```
    assert_that(
        is.factor(data[[group]]),
        length(levels(data[[group]])) >= 2,
        length(unique(data[[group]])) == length(levels(data[[group]])),
        msg = paste(
            "`data[[vars$group]]` must be a factor variable with two or more levels,",
            "all of which must be observed in the data"
        )
    )
```

It didn't look like a consensus was reached here, if you want this condition to be relaxed please can you open an issue so we can discuss some more

PR #528, opened 2025-09-26
527-faster-tests

gravesti:


gravesti:
@gowerc Sure, I don't want to reopen old wounds with those non-reproducible errors. I can start a new PR, but I'm worried that with the current test server performance we can probably only manage one stan model compilation.

gravesti:
Ok, thanks @gowerc . I've blocked some time tonight, so will go through the release plan.

PR #535, opened 2025-11-26
Stop rbmi writing to cache directory on CRAN

### LINKED ISSUE #537: rbmi writing files to cache directory during CRAN tests
rbmi is leaving files in the cache directory when run on CRAN which is causing a note in the CRAN R CMD check logs:

```
Found the following files/directories:
  ‘~/.cache/R/rbmi/rbmi_MMRM_us_default_999123a8603e2aaf4bc799fe64ecafd9.rds’
  ‘~/.cache/R/rbmi/rbmi_MMRM_us_default_999123a8603e2aaf4bc799fe64ecafd9.stan’
* DONE
```

Primary cause appears to be due to rbmi writing `*.stan` files to the cache directory even when `rbmi.enable_cache = FALSE`.  That being said I am confused as to why the `*.rds` file is still being created as Stan should have autowrite disabled during CRAN tests. 

https://www.stats.ox.ac.uk/pub/bdr/donttest/rbmi.out

gravesti:
Closes #537

### DIFF
--- a/tests/testthat.R
+++ b/tests/testthat.R
@@ -1,5 +1,7 @@
 library(testthat)
 library(rbmi)
 
+# Disable cache by default during unit tests
+options("rbmi.enable_cache" = FALSE)
 
 test_check("rbmi")
--- a/tests/testthat/setup.R
+++ b/tests/testthat/setup.R
@@ -8,6 +8,6 @@
 #
 
 # Default cache to being disabled unless running a core or extended tests
-if (!is_core_test()) {
-    options("rbmi.enable_cache" = FALSE)
+if (is_core_test()) {
+    options("rbmi.enable_cache" = TRUE)
 }


R/utilities.R
```diff
@@ -752,6 +711,31 @@ find_stan_file <- function(file, subdir = "") {
...
+#' Get a unique hash
+#'
+#' Gets a unique string on the string content but also including the current session
+#' packages + R version
+#'
+#' @importFrom utils sessionInfo
+#' @keywords internal
+get_unique_hash <- function(content) {
+    pkg_versions <- vapply(
+        sessionInfo(c("rbmi", "rstan", "Rcpp", "RcppEigen"))[["otherPkgs"]],
```
gravesti: I discovered yesterday that `Rcpp` and `RcppEigen` are not strictly required to be installed at runtime, because `stan` only depends on them as `linking-to` (similar to the issues we had with `BH`).

gowerc:
@gravesti  / @danielinteractive  / @luwidmer

Some notes about the changes that I've made here:

- @gravesti  created a workflow that replicated the error, I extracted the core logic of this into a shell script which can in theory be run locally and then updated the workflow to just call that shell script. You can see it replicating the error here: https://github.com/insightsengineering/rbmi/actions/runs/19768301887/job/56646410987

- I updated the cache directory to default to `tempfile()`. This means by default it is unique to just the current R session so you will always have to recompile across sessions. The thinking here is this should significant reduce the risk of future issues as this directory is automatically cleaned up after R has finished and doesn't trigger the `_R_CHECK_THINGS_IN_OTHER_DIRS_` test 
> Currently the directories monitored are the home directory, /tmp (excluding RtmpXXXXXX directories), /dav/shm, ~/.cache (recursively) and ~/.local/share (recursively) or their equivalents on Windows and macOS (the directories in which the default settings for tools::R_user_dir() use a R subdirectory).
https://cran.r-project.org/doc/manuals/r-release/R-ints.html#Tools

- I updated the hashing function to consume the content of the file itself. This means if any line of code changes as well as the R version / R package versions change then a new hash is generated. This way there should be no risk of using an old cached file. As such I have removed the "clean_cache()" function as the logic was getting a bit awkward and I don't think its needed anymore.

- I updated the code so that it constructs the model fully in memory and only ever writes to disk if `getOption("rbmi.enable_cache") == TRUE`

- `options("rbmi.enable_cache")` is set to `FALSE` by default for all tests unless the global environment variable `R_TEST_LOCAL=TRUE`

I will disable all Stan tests in a future PR

gravesti (commented):
@gowerc Looks good, just one note

R/utilities.R
```diff
@@ -752,6 +711,31 @@ find_stan_file <- function(file, subdir = "") {
...
+#' Get unique hash
+#'
+#' Gets a unique string on the string content but also including the current session
+#' packages + R version
+#'
+#' @importFrom utils sessionInfo
+#' @keywords internal
+get_unique_hash <- function(content) {
+    pkg_versions <- vapply(
+        sessionInfo(c("rbmi", "rstan", "Rcpp", "RcppEigen"))[["otherPkgs"]],
```
luwidmer: Why not use `packageVersion()` here?

R/utilities.R
gowerc: Further tweaked in c334ab1a in response to feedback from @luwidmer

R/utilities.R
gowerc: Adjusted in c334ab1a

luwidmer:
@gowerc you mentioned "I will disable all Stan tests in a future PR" - is this just on CRAN, or everywhere?

gowerc:
Hmmm when running this on my local machine I am sometimes getting errors such as:
```
* checking for new files in some other directories ... NOTE
Found the following files/directories:
  ‘/var/folders/6s/f6_p_03n6g182gs7f_fzf_5w0000gn/T/RtmpvPPIP6’
  ‘/var/folders/6s/f6_p_03n6g182gs7f_fzf_5w0000gn/T/RtmpxynGyO’
* DONE
```

My assumption is this is due to other background processes adding folders but I am not confident... in particular the fact the have `Rtmp` in the root file/folder name worries me a bit...

EDIT:

Definitely appears to be R related :(

```
~/Desktop/Work/rbmi donttest* 11s
> cd /var/folders/6s/f6_p_03n6g182gs7f_fzf_5w0000gn/T/RtmpvPPIP6     

/var/folders/6s/f6_p_03n6g182gs7f_fzf_5w0000gn/T/RtmpvPPIP6
> ls
libloc_228_cce4ed05d3cb9016.rds
```

image

gravesti (commented):
@gowerc From my POV this looks good to merge. I can't actually approve since I opened the PR.

gowerc (approved):

PR #566, opened 2026-06-18
`ancova_single()` now return residual variance

munoztd0:
Adding residual variance estimate to ancova_single()

port form `junco`: https://github.com/johnsonandjohnson/junco/issues/368


munoztd0:
@danielinteractive

munoztd0:
@luwidmer what are your view on this ?

munoztd0:
or @tobiasmuetze maybe ?

danielinteractive:
Thanks @munoztd0 , putting on my `rbmi` hat here - can we make this inclusion conditional? And then by default still have the same behavior (without the variance) as currently? I guess that will make it much easier for @luwidmer to accept it into the package.

munoztd0:
> - can we make this inclusion conditional? And then by default still have the same behavior (without the variance) as currently? I guess that will make it much easier for @luwidmer to accept it into the package.

Anything that make us go into the right direction , here it is !

Could you also approve the workflow so that it shows passing tests?

munoztd0:
believe the issue is because of conflicts in roxygen versions

image


But right now rbmi is in 7.3.3 (would you like me to bump it to 8.1.0 in another PR ?)

danielinteractive:
Hi @munoztd0 can you have a look if you can merge the `main` branch into here manually? Let me know if any questions

PR #569, opened 2026-07-07
Analysis function  mmrm

### LINKED ISSUE #370: MMRM analysis function for imputed datasets

munoztd0:
Added `mmrm_analyse()` as a built-in analysis function for use with analyse().
 
It fits a single MMRM (via `mmrm::mmrm()`) across all visits and extracts per-visit treatment contrasts, least square means, and residual variance estimates using `emmeans`. Supports 2+ treatment arms and all covariance structures supported by the `mmrm` package. 

Thus: the `emmeans` package has been moved from `Suggests` to `Imports` accordingly.

fix: https://github.com/johnsonandjohnson/junco/issues/370


tobiasmuetze:
In which cases would an MMRM instead of an ANCOVA fitted to each visit separately be statistically sensible?

danielinteractive:
@tobiasmuetze I am not sure exactly about the motivations for Anova vs. MMRM here, but I know that several companies are using this together with multiple imputation. So there are use cases for this

tobiasmuetze:
Thanks, @danielinteractive. 
In a complete data set, as it is the case after MI, an ANCOVA model performed at each visit provides the same results for the point and variance estimation as an MMRM model with unstructured covariance matrix, provided the same covariates are used and all covariate-by-visit interactions are included. ANCOVA tends to be considerably faster than running an MMRM model jointly for all visits, and ANCOVA uses exact small-sample inference, so confidence intervals and p-values are more appropriate. 
I understand that some companies use an MMRM in this setting, but I think it would be helpful to keep best practices in mind when deciding what to include in {rbmi}.

danielinteractive:
Thanks @tobiasmuetze , fully agree - I would propose to highlight and explain these points in the function documentation such that users are made aware of this.

tobiasmuetze:
Given that `{rbmi}` already supports custom analysis functions, I suggest adding a vignette that demonstrates how to implement an MMRM approach using this functionality rather than adding a separate MMRM analysis function to the package. The vignette could:

- Outline how to define an MMRM approach via a custom analysis function.
- Highlight the settings in which MMRM and ANCOVA produce identical results.
- Discuss the runtime disadvantage of MMRM models compared with ANCOVA alternatives.
- Provide a brief summary of the literature on which factors should be included in the MMRM model (e.g., interactions with visit).

PR #594, opened 2026-09-14
587: Add `rubin_method` option to `pool`

### LINKED ISSUE #587: Add option to use original Rubin (1987) degrees-of-freedom pooling
### Feature Request

Currently, the `rbmi` package uses the Barnard-Rubin (1999) method for pooling degrees of freedom when combining results from multiple imputations (see `rbmi:::rubin_df()`). This is an appropriate default and is statistically preferable in many situations, particularly for small samples.

However, it would be useful to provide an option to use the original Rubin (1987) degrees-of-freedom method as an alternative.

The Barnard-Rubin adjustment should remain the default, with the original Rubin method available as an explicit user-selectable option.

### Motivation

Some legacy analyses (e.g., `SAS PROC MIANALYZE` without the `EDF` option) use the original Rubin pooling approach. Supporting this method in `rbmi` would allow users to:
1. Reproduce historical analyses and validate R results against legacy SAS outputs.
2. Facilitate migration from SAS to R while maintaining consistency with previously generated study outputs.
3. Compare results obtained using the original Rubin and Barnard-Rubin methods when required.

### Details

Image

danielinteractive:
closes #587


R/pool.R
```diff
@@ -483,7 +563,12 @@ rubin_rules <- function(ests, ses, v_com) {
     var_b <- var(ests)
     var_t <- var_w + var_b + var_b / M
 
-    df <- rubin_df(
+    df_fun <- switch(
+        method,
+        modern = rubin_df,
+        original = rubin_orig_df
```
wwojciech, line 569: Hi @danielinteractive - just wondering if `modern` and `original` are the best names  (values) to chose here....

danielinteractive: No probably not the best names, more a placeholder - I only thought about these for about 10 seconds. Happy about suggestions!

wwojciech: Hmm, maybe `"rubin"`, vs `"barnard-rubin"` ?

danielinteractive: Thanks, done 😄

R/pool.R
```diff
@@ -14,6 +14,11 @@
 #' Only used if `method_condmean(type = "bootstrap")` was specified
 #' in the original call to [draws()].
 #'
+#' @param rubin_method a character string specifying the degrees-of-freedom
```
wwojciech: Maybe `rubin_df_method`?

R/pool.R
```diff
@@ -14,6 +14,11 @@
...
+#' approximation used when pooling with Rubin's rules. `"barnard-rubin"` (default)
+#' uses the Barnard-Rubin (1999) small-sample adjustment, while `"rubin"`
+#' uses the approximation from Rubin (1987). Ignored for other pooling methods.
```
wwojciech: Barnard-Rubin (1999) + Rubin (1987): would be add to References section.

R/pool.R
```diff
@@ -25,8 +30,9 @@
 #' call to [draws()]; In particular:
 #'
 #' - `method_approxbayes()` & `method_bayes()` both use Rubin's rules to pool estimates
-#'  and variances across multiple imputed datasets, and the Barnard-Rubin rule to pool
-#'  degree's of freedom; see Little & Rubin (2002).
+#'  and variances across multiple imputed datasets. By default, the Barnard-Rubin
```
wwojciech: Barnard-Rubin (1999)

R/pool.R
```diff
@@ -356,7 +377,15 @@ get_ests_bmlmi <- function(ests, D) {
...
 #' @rdname pool_internal
 #' @export
-pool_internal.rubin <- function(results, conf.level, alternative, type, D) {
+pool_internal.rubin <- function(
+    results,
+    conf.level,
+    alternative,
+    type,
+    D,
+    rubin_method,
```
wwojciech: Here, I would use `df_method` - because this is already the method dedicated for rubin.

R/pool.R
```diff
@@ -372,7 +401,8 @@ pool_internal.rubin <- function(results, conf.level, alternative, type, D) {
     res_rubin <- rubin_rules(
         ests = ests,
         ses = ses,
-        v_com = v_com
+        v_com = v_com,
+        method = rubin_method
```
wwojciech: `df_method = df_method`

R/pool.R
```diff
@@ -464,8 +531,21 @@ rubin_df <- function(v_com, var_b, var_t, M) {
...
+#'
+#' Rubin, D.B. (1987). Multiple Imputation for Nonresponse in Surveys.
+#' John Wiley & Sons, New York. \[Section 3.3\]
 #' @importFrom stats var
-rubin_rules <- function(ests, ses, v_com) {
+rubin_rules <- function(
+    ests,
+    ses,
+    v_com,
+    method
```
wwojciech, line 567: I would definitely rename it to `df_method`.

R/pool.R
```diff
@@ -483,7 +563,12 @@ rubin_rules <- function(ests, ses, v_com) {
     var_b <- var(ests)
     var_t <- var_w + var_b + var_b / M
 
-    df <- rubin_df(
+    df_fun <- switch(
+        method,
+        "barnard-rubin" = rubin_df,
+        "rubin" = rubin_orig_df
```
wwojciech, line 594: I guess, for the backward compatibility it is not possible to have:
```r
"barnard-rubin" = barnard_rubin_df,
 "rubin" = rubin_df
```
?

R/pool.R
```diff
@@ -433,28 +463,65 @@ rubin_df <- function(v_com, var_b, var_t, M) {
...
+#' John Wiley & Sons, New York. \[Section 3.3\]
+rubin_orig_df <- function(v_com, var_b, var_t, M) {
+    var_w <- var_t - (1 + 1 / M) * var_b
+    r <- (1 + 1 / M) * var_b / var_w
+
+    if (r == 0) {
+        return(Inf)
+    }
+
+    return((M - 1) * (1 + 1 / r)^2)
```
wwojciech:
```r
if (r == 0) {
  Inf
} else {
  (M - 1) * (1 + 1 / r)^2
}
```

wwojciech: I think this body might be a bit better:
```r
if (var_b == 0) {
  Inf
} else {
  var_w <- var_t - (1 + 1 / M) * var_b

  # To ensure that `r` is well-defined.
  if (var_w <= 0) {
    stop("Within-imputation variance estimate `var_w` must be positive.")
  }

  r <- (1 + 1 / M) * var_b / var_w
  (M - 1) * (1 + 1 / r)^2
}
```
because:
1. I think mathematically, the idea is that: "if the between variance estimate is 0" then pooled df is Inf.
2. of the numerical stability (since `r` is a floating point quantity) (of course, `var_b` is also a floating-point quantity, but in that case I think it depends on the user, and our function is fine as it is. :)

R/pool.R
```diff
@@ -433,28 +463,65 @@ rubin_df <- function(v_com, var_b, var_t, M) {
...
+#'
+#' @details Let \eqn{V_W = V_T - (1 + 1 / M) V_B} be the within-imputation
+#' variance and \eqn{r = (1 + 1 / M) V_B / V_W} be the relative increase in
+#' variance due to nonresponse. The returned degrees of freedom are
+#' \eqn{(M - 1) (1 + 1 / r)^2}.
+#'
+#' @references
+#' Rubin, D.B. (1987). Multiple Imputation for Nonresponse in Surveys.
+#' John Wiley & Sons, New York. \[Section 3.3\]
+rubin_orig_df <- function(v_com, var_b, var_t, M) {
```
wwojciech, line 501: Why there are no assertions? :)

R/pool.R
```diff
@@ -14,6 +14,11 @@
 #' Only used if `method_condmean(type = "bootstrap")` was specified
 #' in the original call to [draws()].
 #'
+#' @param rubin_method a character string specifying the degrees-of-freedom
+#' approximation used when pooling with Rubin's rules. `"barnard-rubin"` (default)
+#' uses the Barnard-Rubin (1999) small-sample adjustment, while `"rubin"`
```
wwojciech, line 19: Barnard-Rubin (1999): would be good to add to References section, or, "see [rubin_df()] for more details".

Also, since you have a `.bib` file - why not to use `\insertCite{Barnard1999;textual}{rbmi}`.
(maybe `references.bib` needs to be moved to `.\inst` dir., and it think this requires `{Rdpack}`. But this is really nice then :) so maybe as a separate issue to make it for all other references too?

R/pool.R
```diff
@@ -48,6 +54,9 @@
 #' Roderick J. A. Little and Donald B. Rubin. Statistical Analysis with Missing
 #' Data, Second Edition. John Wiley & Sons, Hoboken, New Jersey, 2002. \[Section 5.4\]
 #'
+#' Donald B. Rubin. Multiple Imputation for Nonresponse in Surveys. John Wiley
+#' & Sons, New York, 1987. \[Section 3.3\]
+#'
```
wwojciech, line 62: Since you have a `.bib` file; use `\insertRef{Rubin1987}{rbmi}`

wwojciech (changes requested):
Hi @danielinteractive 
Thank you, overall looks good, some minor comments to consider.

wwojciech (commented):
NA

R/pool.R
```diff
@@ -433,28 +463,65 @@ rubin_df <- function(v_com, var_b, var_t, M) {
...
 
 
+#' @title Original Rubin degrees of freedom approximation
+#'
+#' @description Compute the degrees of freedom according to the original
+#' Rubin (1987) approximation.
+#'
+#' @param v_com Degrees of freedom in the complete-data analysis. Included for
+#' interface compatibility with [rubin_df()] but not used by this approximation.
+#' @param var_b Between-imputation variance of the point estimates.
```
wwojciech: "variance of the point estimates" might be perceived as a bit imprecise, since variance typically refers to a distribution rather than a fixed number, unless you are specifically referring to the sample variance. 

I would make it a bit more precise:

```
#' @param var_b (`numeric(1)`) \cr 
#'   An estimate of the between-imputation variance of the estimator
#'   of the parameter of interest, calculated as the sample variance of the estimates across the imputed datasets.
```

R/pool.R
```diff
@@ -433,28 +463,65 @@ rubin_df <- function(v_com, var_b, var_t, M) {
...
+#' @param var_t Total variance of the point estimate according to Rubin's rules.
```
wwojciech: An estimate of the variance of the pooled estimator of the parameter of interest.

R/pool.R
danielinteractive: ok done

R/pool.R
danielinteractive: For Rdpack that will be a separate issue (https://github.com/openpharma/rbmi/issues/591 is already there actually) and PR since this is orthogonal to these updates.

R/pool.R
danielinteractive: see above

R/pool.R
danielinteractive: I think it is better in this case to carry through the same argument name for easier developer readability. The user won't call this method directly anyway.

R/pool.R
danielinteractive: see above

R/pool.R
wwojciech: ok, so `rubin_df_method` then? :)

R/pool.R
danielinteractive: I think it is ok like this too. Again this function is also not called by the user directly. The other arguments are also briefly named.

R/pool.R
danielinteractive: yes the sample variance is meant, added the word

R/pool.R
danielinteractive: added

R/pool.R
danielinteractive: thanks, done

R/pool.R
danielinteractive: correct, although these functions are not exported I would be careful to now change the meaning of `rubin_df` because people might still have used it.

wwojciech: Since `rubin_df` was not exported, it would actually be legitimate to change its meaning (if someone used it - it was on its own risk). Anyway, just to double consider - as `rubin_df ` / `barnard_rubin_df` is nice to have rather than `rubind_df` / `rubin_orig_df `. If that it is not ok - feel free to ignore the comment :)

wwojciech (approved):
thank you @danielinteractive , please check CI / CD failures if possible.

danielinteractive:
@luwidmer @tobiasmuetze can you have a look from your side too?

gowerc:
I've not gone back to the reference paper to double check that the formula is correct (i'm assuming the code review from @wwojciech has covered that), but otherwise everything looks good here from my perspective:

- documented 👍 
- references 👍 
- not breaking backwards compatibility 👍 
- tested 👍 
- code looks reasonable  👍 

so I'll give it a day or so to see if @luwidmer @tobiasmuetze  have anything to add but if not I will merge.

wwojciech:
> I've not gone back to the reference paper to double check that the formula is correct (i'm assuming the code review from @wwojciech has covered that), but otherwise everything looks good here from my perspective:
> 
> * documented 👍
> * references 👍
> * not breaking backwards compatibility 👍
> * tested 👍
> * code looks reasonable  👍
> 
> so I'll give it a day or so to see if @luwidmer @tobiasmuetze have anything to add but if not I will merge.

Hi @gowerc - I don't have access to the Rubin (1987) book, but I followed the SAS documentation as a reference ( https://documentation.sas.com/doc/en/pgmsascdc/v_079/statug/statug_mianalyze_details08.htm ), and the formula implemented in `rubin_orig_df()` looks correct to me. 

### Minor follow-up

@danielinteractive - one small follow-up thoughts after looking at the PR again:

1. It might be good to add an assertion that `M >= 2`, since for `M = 1` the Rubin df formula is not-defined (because `var_b` is not defined).

2. Making the error message that I initially recommended more precise. For example:
   ```r
   if (var_w <= 0) {
     stop(sprintf(
       "Invalid values of `var_t`, `var_b`, and `M`: ",
       "the implied within-imputation variance `var_w` must be positive; got `var_w = %f`.",
       var_w
     ))
   }
   ```
   The reason is that
   ```r
   var_w <- var_t - (1 + 1 / M) * var_b
   ```
   is not independently estimating `var_w` here. Rather, `var_w` is recovered from the supplied `var_t`, `var_b`, and `M` using Rubin's variance decomposition. Chronologically, `var_t` is obtained upstream from `var_w` and `var_b` (and `M`), whereas the within-imputation variance itself is estimated as: `var_w <- mean(ses^2)`.
   Therefore, a non-positive `var_w` indicates that the supplied `var_t`, `var_b`, and `M` are inconsistent with the variance decomposition, rather than that `var_w` was independently estimated to be non-positive.

3. For a moment I considered simplifying the df formula (for the case of `var_b > 0`) algebraically to

   ```r
   (M - 1) * (var_t / ((1 + 1 / M) * var_b))^2
   ```

   which is mathematically equivalent to the original formula and could be explained in the documentation.
   This could potentially improve numerical stability by avoiding the computation of `var_w` and `r`. However, it would cause that the result is not necessarily exactly in sync with the version that uses `var_w` directly. Since we use `var_w` explicitly to check that the supplied arguments imply a valid within-imputation variance, it would probably be safer to stick with the formula that uses `var_w` directly (as it is now). 
   This is something to consider.

danielinteractive:
Thanks @wwojciech!

I also checked the formula again vs. the SAS documentation and it is correct.

1. Good point, added assertion for `M > 1`
2. Great idea about the algebraic simplification, I simplified it further and aligned the assertion with it. That is now easier to check than the `var_w` thing before.

gowerc (approved):

R/pool.R
```diff
@@ -412,29 +445,81 @@ pool_internal.rubin <- function(results, conf.level, alternative, type, D) {
...
+        fraction <- (M * var_t) / ((M + 1) * var_b)
+
+        if (fraction <= 1) {
+            stop(
+                "(M * var_t) must be larger than ((M + 1) * var_b). ",
+                "Please check var_t, var_b and M."
+            )
         }
+
+        (M - 1) * fraction^2
```
wwojciech, line 519: Thank you @danielinteractive.

I have a concern about finite-precision arithmetic. For the original Rubin df formula to be valid, we need to ensure that the (recovered) `var_w` is positive. The current implementation checks `fraction > 1`, which is algebraically equivalent, but due to finite-precision arithmetic, it is possible for the inequality `fraction > 1` to be satisfied while `var_t - (1 + 1 / M) * var_b <= 0` (or vice versa).

I would therefore suggest implementing the check directly as follows:

```r
adjusted_var_b <- (1 + 1 / M) * var_b

if (var_t - adjusted_var_b <= 0) {
  stop(
    "var_t - (1 + 1 / M) * var_b must be positive. ",
    "Please check var_t, var_b and M."
  )
}

(M - 1) * (var_t / adjusted_var_b)^2
```

This also makes the implementation more directly aligned with the original formulation.

R/pool.R
```diff
@@ -412,29 +445,81 @@ pool_internal.rubin <- function(results, conf.level, alternative, type, D) {
...
+#' \eqn{(M - 1) (1 + 1 / r)^2}. Note that the internal computation is
+#' algebraically simplified to \eqn{(M - 1) ((M V_T) / ((M + 1) V_B))^2}.
+#'
+#' @references
+#' Rubin, D.B. (1987). Multiple Imputation for Nonresponse in Surveys.
+#' John Wiley & Sons, New York. \[Section 3.3\]
+rubin_orig_df <- function(v_com, var_b, var_t, M) {
+    assert_that(is.number(var_b) && noNA(var_b) && var_b >= 0)
+    assert_that(is.number(var_t) && noNA(var_t) && var_t > 0)
+    assert_that(is.count(M) && M > 1)
```
wwojciech, line 505: I find it a bit easier to read `M >= m`, where `m` is the minimum allowed value for `M` (which is `2` in this case). But this is probably just a matter of taste. So, feel free to leave it if you prefer.

wwojciech (changes requested):
Thanks @danielinteractive ! Looks nice overall. I left one minor comment and a second, more important comment regarding finite-precision arithmetic and a possible improvement to address this (if possible, please also styler the code :) ). Thanks!

R/pool.R
danielinteractive: Hmmm... I think it is fine as is - in my version we use the simplest calculation possible here. I am not a numerical expert but I don't see this have any practical implications.

R/pool.R
danielinteractive: Yeah from a programming perspective usually it is better to have `>` than `>=`... but does not matter here anyway

R/pool.R
wwojciech: I’m not an expert on this either, but I’ve encountered cases in my work where, in a very similar setup, finite-prec. arithmetic led to different results. The change doesn’t cost much, but I think it makes the check more robust, so I would vote to update it.

R/pool.R
wwojciech: > Yeah from a programming perspective usually it is better to have `>` than `>=`... but does not matter here anyway

Indeed, I see the point now. Thanks @danielinteractive

R/pool.R
danielinteractive: No I think it is fine to be honest like it is now... and I do think actually that it is numerically more stable... 
so let's just leave it.

wwojciech: My point is that they are evaluated through different sequences of operations, so they can in principle produce different results near the boundary. 

Also, I find the current error message
```r
stop(
  "(M * var_t) must be larger than ((M + 1) * var_b). ",
  "Please check var_t, var_b and M."
)
```
quite enigmatic, as it requires some transformations to understand why this condition must hold. With the formulation using `adjusted_var_b`, it is immediately clear that this is about the recovered `var_w` being positive.

Nevertheless, I’m approving it :)

wwojciech (approved):

R/pool.R
danielinteractive: Yeah but those small differences don't matter in practice. 

To me the current error message is actually easier to understand.

Thanks @wwojciech for the good discussion!

danielinteractive:
Hi @gowerc I think this is ok to merge now.

R/pool.R
wwojciech: Thank you @danielinteractive.

